PR #5, opened 2021-05-20
Gen trainer name

FynnBe:
wandb has the nice functionality of generating run names...  This PR supports this in the default trainer with name=None

### DIFF
--- a/torch_em/trainer/wandb_logger.py
+++ b/torch_em/trainer/wandb_logger.py
@@ -15,12 +15,16 @@ class WandbLogger:
             raise RuntimeError("WandbLogger is not available")
 
         project = os.environ.get("WANDB_PROJECT", None)
-        wandb.init(project=project, config={
-            'name': trainer.name,
+        self.wand_run = wandb.init(
+            project=project,
+            name=trainer.name,
+            # config={
             # 'learning_rate': trainer.learning_rate, # TODO get learning rate from the optimizer
             # TODO parse more of the config from the trainer
-            # ''
-        })
+            # },
+        )
+        if trainer.name is None:
+            trainer.name = self.wand_run.name
 
         self.log_image_interval = trainer.log_image_interval
 
--- a/torch_em/trainer/wandb_logger.py
+++ b/torch_em/trainer/wandb_logger.py
@@ -1,4 +1,6 @@
 import os
+from datetime import datetime
+
 import numpy as np
 
 try:
@@ -23,8 +25,12 @@ class WandbLogger:
             # TODO parse more of the config from the trainer
             # },
         )
+
         if trainer.name is None:
-            trainer.name = self.wand_run.name
+            if os.environ["WANDB_MODE"] == "offline":
+                trainer.name = f"offline-{datetime.now()}"
+            else:
+                trainer.name = self.wand_run.name
 
         self.log_image_interval = trainer.log_image_interval
 
--- a/torch_em/trainer/wandb_logger.py
+++ b/torch_em/trainer/wandb_logger.py
@@ -13,7 +13,7 @@ from .tensorboard_logger import normalize_im, make_grid_image
 
 class WandbLogger:
     def __init__(self, trainer):
-        self.log_dir = f"./logs"
+        self.log_dir = "./logs"
         os.makedirs(self.log_dir, exist_ok=True)
         if wandb is None:
             raise RuntimeError("WandbLogger is not available")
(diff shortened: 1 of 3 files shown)


FynnBe:
not sure if the test will complain about it, too. but locally I get:
```
  File "script.py", line 9, in <module>
    import torch_em.transform
  File "torch-em/torch_em/__init__.py", line 1, in <module>
    from .segmentation import (default_segmentation_loader,
  File "torch-em/torch_em/segmentation.py", line 9, in <module>
    from .trainer import DefaultTrainer
  File "torch-em/torch_em/trainer/__init__.py", line 1, in <module>
    from .default_trainer import DefaultTrainer
  File "torch-em/torch_em/trainer/default_trainer.py", line 11, in <module>
    from .tensorboard_logger import TensorboardLogger
  File "torch-em/torch_em/trainer/tensorboard_logger.py", line 18, in <module>
    from ..loss import EMBEDDING_LOSSES
ImportError: cannot import name 'EMBEDDING_LOSSES' from 'torch_em.loss' (torch-em/torch_em/loss/__init__.py)
```
How was this cyclic import not a problem before?

constantinpape:
> How was this cyclic import not a problem before?
> The only import changes in this PR are `import os` and `from datetime import datetime`^^

That's indeed weird. The tests on `main` are passing and from a quick look the test failures here seem to be unrelated.
Can you find a minimal example to reproduce this?
`python -c "import torch_em.transform"` works for me (at least on `main`).

FynnBe:
> How is my this a problem for my PyCharm!?!

> Can you find a minimal example to reproduce this?

yes `import torch_em` in my PyCharm...

constantinpape:
I cannot reproduce it, `python -c "import torch_em"` works fine for me, both on `main` and on the current version of this PR.

constantinpape:
I also had a quick look and I don't see that there is any import that might be cyclic right now.

FynnBe:
> I cannot reproduce it, python -c "import torch_em" works fine for me, both on main and on the current version of this PR.

thanks for checking, outside pycharm this works for me too... already tried renaming the  `torch_em.utils.test` module to `tst`. some pytest integrated magic hicks up there sometimes... but to no prevail.. anyhow, looks like it's a pycharm only problem, so let's just ignore this for now

torch_em/trainer/wandb_logger.py
```diff
@@ -11,16 +13,27 @@
 
 class WandbLogger:
     def __init__(self, trainer):
+        self.log_dir = f"./logs"
```
constantinpape, line 16: I wouldn't use `f""` strings if there's nothing to insert (my linter would complain about this.)

FynnBe:
a general note. due to you adding `trainer.name` to the log path in the mean time this got a bit bigger by moving `log_dir` to the loggers and not having it in trainer. Not sure how big of a change this really is.

torch_em/trainer/wandb_logger.py
constantinpape: On the other hand, should this be f"./logs/{trainer.name}", or does wandb handle this already internally?

constantinpape (commented):
Too make sure I understand correctly: the idea is that if name is `None` and `wandb` is used, it sets the same checkpoint name as the `wandb` run?
That's nice!
But I think we should add a check in `__init__` that `name is None` is only valid if the wandb logger is used.

torch_em/trainer/wandb_logger.py
FynnBe: I think wandb handles this internally... these files are semi temporary anyway (or downloads of artefacts). the log is really just valuable/valid if uploaded to a server (this server might run locally in docker though)
because
```python
run = wandb.init(name=None, dir="logs")
 name = run.name
 ```
 
 I left out the name in the log dir (the default logdir is `./wandb` and that works just fine...

constantinpape: > I think wandb handles this internally... these files are semi temporary anyway (or downloads of artefacts). the log is really just valuable/valid if uploaded to a server (this server might run locally in docker though)

Ok, makes sense.

constantinpape:
> a general note. due to you adding `trainer.name` to the log path in the mean time this got a bit bigger by moving `log_dir` to the loggers and not having it in trainer. Not sure how big of a change this really is.

I added the `name` to properly separate the runs in tensorboard (before they were all mashed together).
If this is not necessary for wandb (as we assume), then moving the `log_dir` to the loggers makes total sense.

PR #8, opened 2021-05-25
add a generic Tile transformation; like numpy.tile

FynnBe:
Not sure about the preferred way to add a dimension agnostic transformation.

This `class Tile(torch.nn.Module)` implementation does not inherite from a kornia class (as there is no public common base class for 2D/3D), but is compatible with them.. maybe there should be an abstract base class for these cases...

### DIFF
--- a/environment_cpu.yaml
+++ b/environment_cpu.yaml
@@ -9,7 +9,6 @@ dependencies:
     - cpuonly
     - elf
     - pip
-    - pytest
     - pytorch
     - tensorboard
     - tifffile
--- a/environment_cpu.yaml
+++ b/environment_cpu.yaml
@@ -5,10 +5,11 @@ channels:
 name:
     torch-em-cpu
 dependencies:
+    - affogato
     - cpuonly
     - elf
-    - affogato
     - pip
+    - pytest
     - pytorch
     - tensorboard
     - tifffile
--- a/environment_gpu.yaml
+++ b/environment_gpu.yaml
@@ -9,7 +9,6 @@ dependencies:
     - cudatoolkit=10.2  # set correct cuda version for your system
     - elf
     - pip
-    - pytest
     - pytorch
     - tensorboard
     - tifffile
--- a/environment_gpu.yaml
+++ b/environment_gpu.yaml
@@ -5,14 +5,15 @@ channels:
 name:
     torch-em
 dependencies:
+    - affogato
     - cudatoolkit=10.2  # set correct cuda version for your system
     - elf
-    - affogato
     - pip
+    - pytest
     - pytorch
     - tensorboard
-    - torchvision
     - tifffile
+    - torchvision
     - tqdm
     - pip:
         - kornia
--- a/test/transform/test_generic.py
+++ b/test/transform/test_generic.py
@@ -1,0 +1,28 @@
+import numpy
+import torch
+import pytest
+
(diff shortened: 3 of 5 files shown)


environment_cpu.yaml
```diff
@@ -5,10 +5,11 @@ channels:
 name:
     torch-em-cpu
 dependencies:
+    - affogato
     - cpuonly
     - elf
-    - affogato
     - pip
+    - pytest
```
constantinpape, line 12: I would rather not mix testing frameworks, can you use unittest?

constantinpape:
> This `class Tile(torch.nn.Module)` implementation does not inherite from a kornia class (as there is no public common base class for 2D/3D), but is compatible with them.. maybe there should be an abstract base class for these cases...

Could you give an example of how you use the `Tile` transform, and how you pass it to the corresponding dataset?
I think that would help to decide whether to add a base class for this.

FynnBe:
> Could you give an example of how you use the Tile transform, and how you pass it to the corresponding dataset?
I think that would help to decide whether to add a base class for this.

Initially I wanted something like:
```python
    raw_transform = torch_em.transform.get_raw_transform(
        augmentation1=torch_em.transform.EMDefectAugmentation(p_drop_slice=p_drop_slice),
        augmentation2=torch_em.transform.Tile(reps=(2, 1, 1, 1, 1), match_shape_exactly=False),
    )
    label_transform = torch_em.transform.Tile(reps=(2, 1, 1, 1, 1), match_shape_exactly=False)  # todo: expand/braodcast_to instead of tile

    loader = torch_em.default_segmentation_loader(
        ...,
        raw_transform=raw_transform,
        label_transform=label_transform,
        transform=...,
        ...
)
```
-> creating two versions of the same input due to independent transformations following `Tile`

But I think I'll do something more explicit/manual instead... Still thought the `Tile` could come in handy some time.

FynnBe:
Instead of a general transformation base class a [runtime checkable protocol](https://docs.python.org/3/library/typing.html#typing.runtime_checkable) would probably the proper way to go to be compatible with kornia, without inheriting from the protected `kornia.augmentation.base._AugmentationBase`

constantinpape:
I don't really understand the issue though. Neither raw transforms nor label transforms must inherit from kornia augmentations. The only requirement for them is that they are callable and take a single argument (raw data /  labels).

FynnBe:
Ah, I see. So how do you combine transformations? I thought this should be done with the `KorniaAugmentationPipeline` , but that failed due to it acting on multiple tensors rather than an indiviudal one. So if I want to mix/concatenate transforms/augmentation I should just make a callable transform class and put everything in there, that I need?

constantinpape:
> Ah, I see. So how do you combine transformations?

Chaining callables is so trivial that I haven't implemented anything special for this and either wrote a small local function or used functools.reduce.
But we could add something like `Chain` or `Concatenate` to the generic transformations.

> I thought this should be done with the `KorniaAugmentationPipeline`

That's only necessary for kornia augmenations, because it does several fancy things to manage the random state of them etc.

FynnBe:
I think I just tried too hard to use what's already there^^
It would also be easy to write a wrapper around a simple transform to make it kornia pipeline compatible... 
We can leave this for future PRs

constantinpape (approved):

PR #9, opened 2021-05-25
Wandb logger

FynnBe:
Improve offline wandb run name (no colon). Now the resulting file path is also valid on windows.

hitchhiking on this PR: `import torch.utils.data` for `torch.utils.data.DataLoader`  (previously it was only `import torch`)

### DIFF
--- a/torch_em/segmentation.py
+++ b/torch_em/segmentation.py
@@ -1,7 +1,7 @@
 import os
 from glob import glob
 
-import torch
+import torch.utils.data
 from elf.io import open_file
 
 from .data import ConcatDataset, ImageCollectionDataset, SegmentationDataset
--- a/torch_em/segmentation.py
+++ b/torch_em/segmentation.py
@@ -1,6 +1,7 @@
 import os
 from glob import glob
 
+import torch
 import torch.utils.data
 from elf.io import open_file
 
--- a/torch_em/trainer/wandb_logger.py
+++ b/torch_em/trainer/wandb_logger.py
@@ -31,7 +31,7 @@ class WandbLogger:
 
         if trainer.name is None:
             if os.environ.get("WANDB_MODE") == "offline":
-                trainer.name = f"offline-{datetime.now()}"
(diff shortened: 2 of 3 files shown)


torch_em/segmentation.py
```diff
@@ -1,7 +1,7 @@
 import os
 from glob import glob
 
-import torch
+import torch.utils.data
```
constantinpape, line 4: For me this is sufficient and this works:
```python
import torch
loader = torch.utils.data.DataLoader
```
Maybe depends on the torch version?

FynnBe: oh, true, I was misled by PyCharm complaining. I think there is some lazy module loading involved for torch.utils. If you don't mind I would extend it for static type checks. 

utils magic in `torch.__init__.py`:
```python
...
from ._utils import _import_dotted_name
from ._utils_internal import get_file_path, prepare_multiprocessing_environment, \
    USE_RTLD_GLOBAL_WITH_LIBTORCH, USE_GLOBAL_DEPS
    ...
```

constantinpape: But we use `torch` in several other places, e.g. https://github.com/constantinpape/torch-em/blob/main/torch_em/segmentation.py#L189.
So this change would break several things.

FynnBe: I don't see this breaking anything..
```python
>>> import torch.utils.data
>>> torch.float32
torch.float32
```
I mean torch.__init__ has to execute when importing `torch.<anything>`, or am I missing some fundamental python quirks?

constantinpape: Yeah, you're right this still works (though I think it's possible to write mean things that would make this not work). It's just conceptually weird to ask for a specific torch module when you use functionality from the top-level one.

FynnBe, line 4: nothing stopping us from importing both
```suggestion
import torch
import torch.utils.data
```

constantinpape (approved):

PR #10, opened 2021-05-28
Invertable affine augmentations

FynnBe:
idea: the augmentation pipleine gets an additional kwarg: return_transform (default False) just like all kornia augmentations have. If True it returns a the transformed tensors and the responding affine transformation matrices. These can be used, e.g. to call `AugmentaionPipeline.apply_inverse` to invert the (geometric part of the) transformations.


torch_em/transform/augmentation.py
```diff
@@ -1,6 +1,11 @@
-import torch
+from typing import List, Optional, Sequence, Tuple, Union
+
+import kornia.augmentation.utils
 import numpy as np
-import kornia
```
constantinpape, line 3: Can we leave the single import please instead of splitting it into 5 different ones?

torch_em/transform/augmentation.py
```diff
@@ -9,18 +14,18 @@
 # TODO RandomElastic3D ?
 
 
-class RandomElasticDeformation(kornia.augmentation.AugmentationBase2D):
```
constantinpape, line 12: Please don't make the format changes,that makes it really hard to review when things are more complex.

torch_em/transform/augmentation.py
```diff
@@ -63,13 +68,18 @@ def apply_transform(self, input, params):
 
 # TODO implement 'require_halo', and estimate the halo from the transformations
 # so that we can load a bigger block and cut it away
-class KorniaAugmentationPipeline(torch.nn.Module):
-    interpolatable_torch_tpyes = [torch.float16, torch.float32, torch.float64]
-    interpolatable_numpy_types = [np.dtype('float32'), np.dtype('float64')]
+class AugmentationPipeline(torch.nn.Module):
```
constantinpape, line 71: I think "KorniaAugmentationPipeline" makes more sense as a name, because it's not for generic Augmentations.

torch_em/transform/augmentation.py
```diff
@@ -90,32 +100,72 @@ def is_interpolatable(self, tensor):
...
+                    assert all_params[ai] == aug._params
+
+            transformed_tensors.append(tensor)
+
+        return (transformed_tensors, trans_matrices) if self.return_transform else transformed_tensors
 
     def halo(self, shape):
         return self.halo
 
+    def apply_inverse(self, *tensors: torch.Tensor, forward_transforms: Sequence[torch.Tensor], padding_mode="border"):
```
constantinpape, line 155: This only works for affine trafos / anything that can be expressed as affine trafo. But that's certainly not the case for all augmentations we have, e.g. elastic deformations.

constantinpape (commented):
> the augmentation pipleine gets an additional kwarg: return_transform (default False) just like all kornia augmentations have.

I think that part makes sense.

> If True it returns a the transformed tensors and the responding affine transformation matrices.

But the transformation is in general not an affine matrix. Quite a few augmentations can be expressed as affine, but this does not hold true in general, e.g. for elastic deformations.

> These can be used, e.g. to call `AugmentaionPipeline.apply_inverse` to invert the (geometric part of the) transformations.

I think we need to be aware of what transformation and then choose the correct inverse function based on the transformation (or throw some error if the inverse for the transformation is not defined.)

torch_em/transform/augmentation.py
FynnBe: renamed it to `apply_inverse_affine` to make it explicit that it applies only to transformations that can be expressed as an affine transformation.

constantinpape:
I think it would still make sense to eventually come up with a more general functionality for inverting the augmentations, but for now we can go with this option. I think we should def. add tests here though...

PR #12, opened 2021-06-24
Updated SegmentationDataset to support 4D images

JoOkuma:
This PR is related to constantinpape/elf#36

I think using greater or equal could be a compromise. It allows using higher-dimensional datasets, when performing the slicing it will still have the desirable property of crashing when indexing a lower-dimensional dataset with a slice of greater dimensions, and the assertion catches when using a `patch_shape` smaller than the `raw.ndim`, which could lead to silent undesirable behavior.

Let me know if you have any suggestions, cheers.


constantinpape:
Hey @JoOkuma,
sorry for taking so long to come back to you about this. I have implemented support for 4d data now in #42, so I will go ahead and close it.
I will send you example code for how to use this soon.

PR #34, opened 2022-02-11
add boolean to set ilastik threads to 0

JonasHell:
I added the boolean we discussed. Works correctly in my code and boosts the speed almost 5x


constantinpape (approved):
Thanks @JonasHell! The changes look good.
I will also look into adding tests for the shallow2deep part so that CI can find some basic issues.

PR #35, opened 2022-02-14
add plt functionality for check_loader

JonasHell:
you mentioned once that this could be helpful


constantinpape:
Thanks! That's indeed super useful for x-forwarding.

PR #52, opened 2022-04-06
Expand tests for trainer

constantinpape:
TODO
- [ ] more checks for loading a trainer from checkpoint
- [ ] add test for spoco trainer, including loading from checkpoint
- [ ] (probably) update spoco trainer to be compatible with refactor in #51


constantinpape:
This is outdated.

PR #53, opened 2022-04-06
refactor DefaultTrainer.from_checkpoint

FynnBe:


constantinpape:
Only had a brief look at the changes, I think this should work in principle.
But tests are still failing: 
```
  File "/home/runner/work/torch-em/torch-em/test/trainer/test_default_trainer.py", line 81, in test_from_checkpoint
    trainer2 = DefaultTrainer.from_checkpoint(
  File "/home/runner/work/torch-em/torch-em/torch_em/trainer/default_trainer.py", line 195, in from_checkpoint
    trainer_kwargs = cls._get_trainer_kwargs(load)
  File "/home/runner/work/torch-em/torch-em/torch_em/trainer/default_trainer.py", line 175, in _get_trainer_kwargs
    train_loader=load("train_loader"),
  File "/home/runner/work/torch-em/torch-em/torch_em/trainer/default_trainer.py", line 126, in __call__
    return self.load_loader(name)
  File "/home/runner/work/torch-em/torch-em/torch_em/trainer/default_trainer.py", line 154, in load_loader
    ds = self.init_data[f"{name}_dataset"]
KeyError: 'train_loader_dataset'
```
Please make sure that they pass.

torch_em/trainer/default_trainer.py
```diff
@@ -76,64 +79,119 @@ def iteration(self):
...
+            self.save_path = save_path
+            self.device = torch.device(self.init_data["device"]) if device is None else torch.device(device)
+
+        def __call__(
+            self,
+            name: str,
+            *dynamic_args,
+            optional=False,
+            only_class=False,
+            dynamic_kwargs: Optional[Dict[str, Any]] = None,
```
constantinpape, line 121: Is there a particular reason you're not using `**dynamic_kwargs` here?

torch_em/trainer/default_trainer.py
```diff
@@ -76,64 +79,119 @@ def iteration(self):
...
+        ):
+            if name.endswith("_loader"):
```
constantinpape, line 123: Can you please explain this?

torch_em/trainer/default_trainer.py
```diff
@@ -76,64 +79,119 @@ def iteration(self):
...
+        """
+
+        def __init__(self, init_data: dict, save_path: str, device: Union[str, torch.device]):
+            self.init_data = init_data
+            self.save_path = save_path
+            self.device = torch.device(self.init_data["device"]) if device is None else torch.device(device)
+
+        def __call__(
+            self,
+            name: str,
```
constantinpape, line 117: What exactly is `name` here? I thought it was the name of the trainer, i.e. the checkpoint name. But looks like it's something else.

constantinpape (commented):
Looks good overall, but two minor comments left to address.

torch_em/trainer/default_trainer.py
constantinpape: Ok, I understand now that this is the `name` of the `kwarg`. But this should be explained a bit better, it's somehow not immediately obvious. I think it would be best if you can just add the "normal usage" in the doc string, i.e. how the deserializer is normally called for a kwarg.

torch_em/trainer/default_trainer.py
FynnBe: yes, `**dynamic_kwargs` would not allow to `optional` (or `only_class`) as a kwarg

torch_em/trainer/default_trainer.py
FynnBe: good idea. I suppose we can additionally rename `name` to `kwarg_name` or `attribute`?

constantinpape: Yes, let's go with `kwarg_name`

torch_em/trainer/default_trainer.py
constantinpape: Maybe I am missing something, but I don't see why this is the case. As far as I can see this is perfectly valid python:
```python
def my_func(
    kwarg_name, *args,
    optional=False, only_class=False,
    **kwargs
):
    print(optional)

my_func("x", "y", optional=True, blub="xyz")
```
(prints True)

constantinpape (commented):
I still don't see why we can't use `**kwargs` in `__call__`. Looks good otherwise.

FynnBe:
> I still don't see why we can't use `**kwargs` in `__call__`. Looks good otherwise.

This would fail in case someone needs a special dynamic_kwarg that happens to also be called `optional`... 
It's an edge case, sure...
The updated, simplified version of this does not have this issue anymore: https://github.com/FynnBe/torch-em/pull/1
It builds on top of this PR, but changes the Deserializer a bit to be more analogous to the new Serializer, which is used in `_build_init`...

constantinpape:
Ah ok, I see what you mean. Quite a corner case indeed, but good to catch it here.

> The updated, simplified version of this does not have this issue anymore: [FynnBe#1](https://github.com/FynnBe/torch-em/pull/1)

Cool, I will have a look once you think it's good to go and create it as a PR here, for now I think the changes in this one are good to go and I'll merge.

PR #54, opened 2022-04-08
Add Serializer to DefaultTrainer 

FynnBe:
also changes the Deserializer a bit so it is more obvious how they function in combination...

I didn't want to move code too much, but maybe after this PR the DefaultTrainer could do with some clean-up, probably moving the definitions of Deserializer and Serializer closer together would be helpful.


torch_em/trainer/default_trainer.py
```diff
@@ -158,93 +160,173 @@ def load_generic(
...
-        if self.lr_scheduler is not None:
-            init_data["lr_scheduler_class"] = _full_class_path(self.lr_scheduler)
-            init_data["lr_scheduler_kwargs"] = get_constructor_arguments(self.lr_scheduler)
-        return init_data
+    class Serializer:
+        """Implements how to serialize trainer kwargs from a trainer instance
+
+        Examples:
+            To extend the serialization process you can inherite from this Serializer in a derived Trainer class.
+            Note that `DefaultTrainer.Serializer.dump_generic()` covers most cases already.
```
constantinpape, line 207: There is no `dump_generic` method, but different specializations of this. I suggest to change the text here a bit, otherwise it's a bit confusing that this refers to a method that doesn't exist.

constantinpape (commented):
Only one minor comment, otherwise looks good to go.
And I agree that some clean up would be nice, but should be done in a follow-up PR.

torch_em/trainer/default_trainer.py
FynnBe: fixed in https://github.com/constantinpape/torch-em/pull/54/commits/68887b1125c2c80a422737780057ea93e7863b9f#

PR #55, opened 2022-04-13
Deserialize ROI Wrapper WIP

constantinpape:
cc @JonasHell


constantinpape:
Thanks for making me aware of this @JonasHell and it works now correctly (and is also tested ;))

PR #60, opened 2022-06-03
Add sampler that checks if there are a minimum number of instances in the volume

JonasHell:
When there is only one instance in the volume, the RF is trained with only one class. This leads to an assertion error in `_predict_rf()` in `shallow2deep_dataset.py`. By applying one of the new samplers this issue can be fixed. Therefore, I added the sampler as an argument to `prepare_shallow2deep()`, while keeping the default behavior of having the `MinForegroundSampler`.


constantinpape (approved):
Thanks!

PR #62, opened 2022-06-14
Add functionality to mask to-bg-boundaries

JonasHell:
The new BoundaryTransform labels the to-bg-boundaries with a dedicated label (e.g. -1).
This can already be interpreted by the RF training procedure.
MaskIgnoreLabel in combination with LossWrapper now also allows trainer instances to interpret this correctly.
The LossWrapper was modified to also allow a mask to be given directly (used via ApplyMask).
The new sampler is needed to make sure that there is still foreground signal left in the image after masking the to-bg-boundaries. Unfortunately, right now it calls the new BoundaryTransform.


constantinpape (approved):
Great!

PR #68, opened 2022-07-07
add noise augmentations for mean teacher approach

JonasHell:


torch_em/transform/raw.py
```diff
@@ -36,10 +38,74 @@ def normalize_percentile(raw, lower=1.0, upper=99.0, axis=None, eps=1e-7):
     return normalize(raw, v_lower, v_upper, eps=eps)
 
 
+def normalize_torch(tensor, minval=None, maxval=None, eps=1e-7):
```
constantinpape, line 41: Can we instead update `normalize` so that it also supports torch tensors? I think that the only thing that needs to be adapted is that the arguments to min and max need to be passed as normal arguments, not as kwargs.

constantinpape (commented):
Noise augmentation look good, but I would rather not duplicate the `normalize` fucntion and it should be easy to make it work for both torch and numpy.

torch_em/transform/raw.py
JonasHell: The kwargs are not the only problem, casting is also different and in general min()/max() behaves differently for numpy and torch. But anyway, I updated it now, so normalize() should now work for both.

constantinpape: Thanks!

PR #78, opened 2022-07-22
fix kwargs, other strategies have no img_shape

JonasHell:


constantinpape:
This is already solved by #75 now (by adding `**kwargs` to the other sampling functions)

PR #83, opened 2022-08-03
adapt _serialize_feature_config() to work with strings

JonasHell:
I think there is no additional check needed since `_apply_filters()` already checks if a function with this name exists, and it is called before `_serialize_feature_config()`


constantinpape (approved):
Good point re check and thanks for the PR.

PR #87, opened 2022-09-08
Update spoco losses

constantinpape:
TODO:
- add spoco experiment in `experiments/spoco` and validate that this reproduces the original spoco experiments (try with dsb)
- clarify the different spoco losses and when to use them
  - `ExtendedContrastiveLoss`
  - `SpocoLoss`
  - `SPOCOConsistencyLoss`


constantinpape:
While not everything is done here, this PR has picked up some other very relevant stuff, so I will merge it now and do the spoco checks in a follow-up.

PR #90, opened 2022-09-19
implement gradient masking with cropping

JonasHell:
current state of the gradient masking with cropping.

TODO: allow for different mask in each channel
TODO: make it more efficient

### DIFF
--- a/test/loss/test_loss_wrapper.py
+++ b/test/loss/test_loss_wrapper.py
@@ -41,7 +41,7 @@ class TestLossWrapper(unittest.TestCase):
         t = torch.rand(*shape)
         m = torch.rand(*shape) > .5
         p_masked, t_masked = ApplyMask()(p, t, m)
-        out_shape = (m.sum(), 1)
+        out_shape = (m.sum(), shape[1])
         self.assertTrue(p_masked.shape == out_shape)
         self.assertTrue(t_masked.shape == out_shape)
 
@@ -51,7 +51,7 @@ class TestLossWrapper(unittest.TestCase):
         t = torch.rand(*shape)
         m = torch.rand(*shape) > .5
         p_masked, t_masked = ApplyMask()(p, t, m)
-        out_shape = (m.sum(), 1)
+        out_shape = (m.sum(), shape[1])
         self.assertTrue(p_masked.shape == out_shape)
         self.assertTrue(t_masked.shape == out_shape)
 
@@ -62,6 +62,17 @@ class TestLossWrapper(unittest.TestCase):
         m = torch.rand(*shape) > .5
         with self.assertRaises(ValueError):
             p_masked, t_masked = ApplyMask()(p, t, m)
+
+        # _crop different shapes
+        shape_pt = (5, 2, 10, 128, 128)
+        p = torch.rand(*shape_pt)
+        t = torch.rand(*shape_pt)
+        shape_m = (5, 1, 10, 128, 128)
+        m = torch.rand(*shape_m) > .5
+        p_masked, t_masked = ApplyMask()(p, t, m)
+        out_shape = (m.sum(), shape_pt[1])
+        self.assertTrue(p_masked.shape == out_shape)
+        self.assertTrue(t_masked.shape == out_shape)
 
     def test_ApplyMask_output_shape_multiply(self):
         from torch_em.loss import ApplyMask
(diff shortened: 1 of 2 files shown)


torch_em/loss/wrapper.py
```diff
@@ -43,30 +44,72 @@ def forward(self, prediction, target, **kwargs):
...
+            "channel_dim": channel_dim,
+        }
+
+    def __call__(self, prediction, target, mask):
+        mask.requires_grad = False
+        return self.masking_func(prediction, target, mask, self.channel_dim)
+
 
-class ApplyAndRemoveMask:
+class ApplyAndRemoveMask(ApplyMask):
```
constantinpape, line 94: +1 for the inheritance (here and in the other classes), but I think we should then also introduce a `__init__` here that calls the super constructor.

torch_em/loss/wrapper.py
```diff
@@ -43,30 +44,72 @@ def forward(self, prediction, target, **kwargs):
...
-    def __call__(self, prediction, target, mask):
-        mask.requires_grad = False
+    def _crop(prediction, target, mask, channel_dim):
+        mask = mask.type(torch.bool)
+        pred = []
+        trgt = []
+        n_channels = prediction.shape[channel_dim]
+        # for now only support same mask for all channels
+        # to be able to stack/concat afterwards
+        # TODO: make this more efficient, and for different
```
constantinpape, line 54: For the case where we have a mask of lower dimensionality this can be done much simpler (I am using numpy here for simplicity, but the same works with torch, I just checked):
```
x = np.random.rand(128, 128, 3)
y = np.random.rand(128, 128) > 0.5
z = x[y]
print(z.shape)
(8186, 3)  # 8186 is the number of elems in mask, i.e. it's not always the same number
```
So you just need to remove the singleton channel dimension in the mask, swap the channel dimension to the back in the prediction / target, apply the mask and bring the channel dimension back to the original pos.

I still need to think about the case where we have a full mask.

torch_em/loss/wrapper.py
JonasHell: do you mean for better readibility? Because __init__ of the base class is called anyway if we do not overwrite it.

torch_em/loss/wrapper.py
JonasHell: True, I implement it. Thank you :)

torch_em/loss/wrapper.py
constantinpape: Yea, you're right; I didn't think about this one long enough ;).

torch_em/loss/wrapper.py
JonasHell: done

torch_em/loss/wrapper.py
```diff
@@ -43,30 +44,65 @@ def forward(self, prediction, target, **kwargs):
 
 
 class ApplyMask:
-    def __call__(self, prediction, target, mask):
-        mask.requires_grad = False
+    def _crop(prediction, target, mask, channel_dim):
+        mask = mask.type(torch.bool)
+        # TODO: make this work for different
+        # masks per channel
+        assert mask.shape[channel_dim] == 1
```
constantinpape, line 51: The case where we have a full mask is complicated and I can't see an option right now to make it work where we also preserve the N x C structure. I think we can go ahead with the current solution, but should throw a more meaningful error message here. I suggest to throw a `ValueError`, which should also indicate that using `masking_method=multiply` will fix the issue.

constantinpape (commented):
Looks good now, we should improve the error that is being thrown when the mask is not a singleton. Also, could you add a test for `ApplyMask` here https://github.com/constantinpape/torch-em/blob/main/test/loss/test_loss_wrapper.py that checks that the shape after calling it is correct? (N x C x Flattened Spatial).

torch_em/loss/wrapper.py
JonasHell: done

test/loss/test_loss_wrapper.py
```diff
@@ -32,6 +32,50 @@ def test_masking(self):
         # print((grad[~mask] == 0).sum())
         self.assertTrue((grad[~mask] == 0).all())
 
+    def test_ApplyMask_output_shape_crop(self):
```
constantinpape, line 35: Can you add a test where the number of channels in `p` and `t` is > 1, but the mask only has a single channel?

JonasHell: done

PR #93, opened 2022-09-20
Use tifffile to read compressed tif files 

### LINKED ISSUE #92: Add support for multichannel compresed tif files
Currently `torch-em` support only reading files using `tifffile` only if could use `memmap` to read files

https://github.com/constantinpape/torch-em/blob/1487015f24cafdbc6e0079e99034e274d597cfcb/torch_em/util/image.py#L27-L32

but `memmap` could not be used if the file is saved with compression (from tifffile docs):
```
Memory-mapping requires the image data stored in native byte order,
    without tiling, compression, predictors, etc
```

the fallback is to use `imageio` that read-only 2d single channel data. Why not use `tifffile.imread` if the extension is `tif`, `tiff`, `TIF`, `TIFF`?  

Are you open to contributions?

Czaki:
closes #92

I have added test but not executed them. I do not see why test are not executed on pull requests. Maybe you need to approve the run.

### DIFF
--- a/test/util/test_imageio.py
+++ b/test/util/test_imageio.py
@@ -11,7 +11,7 @@ class TestImageRead(unittest.TestCase):
         from torch_em.util.image import load_image, supports_memmap
 
         with tempfile.TemporaryDirectory() as td:
-            tifffile.imwrite(os.path.join(td, "test.tif"), np.zeros(10, 10, 2))
+            tifffile.imwrite(os.path.join(td, "test.tif"), np.zeros((10, 10, 2)))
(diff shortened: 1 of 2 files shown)


constantinpape:
> Maybe you need to approve the run.

Yes, for first time contributors the tests need to be approved (it's a new security feature by github).
I have approved it and it's running now, will check the changes now as well.

Czaki:
> Yes, for first time contributors the tests need to be approved (it's a new security feature by github).

In the past I see messages that the project maintainer need to approve tests to run, and now I do not see such a message so I was confused.

constantinpape (approved):
Thanks for the contribution!

PR #101, opened 2022-12-10
implement gradient masking with cropping - new PR

JonasHell:
I cleaned my git repo. Unfortunately, renaming of branches leads to the closing of connected PRs. So I opened a new one. To not loose the old discussion, I link the old PR [here](https://github.com/constantinpape/torch-em/pull/90).

### DIFF
--- a/test/loss/test_loss_wrapper.py
+++ b/test/loss/test_loss_wrapper.py
@@ -41,7 +41,7 @@ class TestLossWrapper(unittest.TestCase):
         t = torch.rand(*shape)
         m = torch.rand(*shape) > .5
         p_masked, t_masked = ApplyMask()(p, t, m)
-        out_shape = (m.sum(), 1)
+        out_shape = (m.sum(), shape[1])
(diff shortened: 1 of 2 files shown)


constantinpape:
Thanks @JonasHell. I will have a look soon.

constantinpape (commented):
Thanks for working on this! The changes to `loss_wrapper.py` all look good to me.
It would be good to extend the test cases a bit to cover all the changes:
- In `test_masking`: test for both `crop` and `multiply` masking methods that the masked gradients are zero. (Can just put this in a loop where `ApplyAndRemoveMask` is called with the corresponding argument.) The test could also be renamed to `test_ApplyAndRemove...` to make the naming more consistent.
- Add a test for `MaskIgnoreLabel` along the lines of `test_masking`.

JonasHell:
Done :)
Let me know if you want to have more tests.

constantinpape (approved):
The tests look good now. Thanks!

PR #104, opened 2022-12-15
Add Separate Label Path to the LiveCELL Loader

anwai98:


torch_em/data/datasets/livecell.py
```diff
@@ -16,7 +16,7 @@
               "LIVECell/livecell_coco_train.json"),
     "val": ("http://livecell-dataset.s3.eu-central-1.amazonaws.com/LIVECell_dataset_2021/annotations/"
             "LIVECell/livecell_coco_val.json"),
-    "test": ("http://livecell-dataset.s3.eu-central-1.amazonaws.com/LIVECell_dataset_2021/annotations/"
+    "test": ("http://livecell-dataset.s3.eu-central-1.amazonaws.com/LIVECell_dataset_2021/annotations/",
```
constantinpape, line 19: Please don't add this comma, it makes the address wrong (it's a tuple when you add the comma, rather than a multi-line string.)

torch_em/data/datasets/livecell.py
```diff
@@ -143,19 +145,23 @@ def _livecell_segmentation_loader(
                                               label_dtype=label_dtype,
                                               transform=transform,
                                               n_samples=n_samples)
-    return torch_em.segmentation.get_data_loader(ds, batch_size, **loader_kwargs)
+
```
constantinpape, line 148: This change is not necessary. `get_data_loader` already does the right thing: https://github.com/constantinpape/torch-em/blob/main/torch_em/segmentation.py#L302

constantinpape (commented):
Thanks! The addition of `label_path` looks good, but there are two unrelated changes that should be undone.

anwai98:
Updated fixes along all the suggestions mentioned.

PR #108, opened 2023-02-22
add args to adjust tqdm output

JonasHell:
I am cleaning up my code. At some point I added these additional args to adjust the tqdm output because I called this function repeatedly and this resulted in an ugly output. However, I am not sure if you want to keep the changes. Just wanted to make sure they are documented somewhere.


constantinpape (approved):
This is good, thanks!

PR #115, opened 2023-03-28
Update Prob-DA Source UNet Parameters

anwai98:
- updated the patch shape : (512, 512) -> (256,256)
- added `save_root` argument in the `evaluate_source_model` function to route the checkpoints from desired path

### DIFF
--- a/experiments/probabilistic_domain_adaptation/livecell/common.py
+++ b/experiments/probabilistic_domain_adaptation/livecell/common.py
@@ -191,7 +191,7 @@ def _get_image_paths(args, split, cell_type):
 
 
 def get_unsupervised_loader(args, split, cell_type, teacher_augmentation, student_augmentation):
-    patch_shape = (512, 512)
+    patch_shape = (256, 256)
 
     def _parse_aug(aug):
         if aug == "weak":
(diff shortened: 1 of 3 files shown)


experiments/probabilistic_domain_adaptation/livecell/common.py
```diff
@@ -212,11 +215,11 @@ def _parse_aug(aug):
     return loader
 
 
-def get_supervised_loader(args, split, cell_type):
-    patch_shape = (512, 512)
```
constantinpape, line 216: Let's change this in the unsupervised loader as well.

experiments/probabilistic_domain_adaptation/livecell/common.py
```diff
@@ -212,11 +215,11 @@ def _parse_aug(aug):
...
+def get_supervised_loader(args, split, cell_type, batch_size):
```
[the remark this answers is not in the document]
anwai98: Updated it just now in the `unet_adamt.py` training script.

experiments/probabilistic_domain_adaptation/livecell/common.py
anwai98: Updated this as well (however, I am really curious to see how the joint training setup works with more context - (512, 512))

PR #116, opened 2023-03-30
Update Mean-Teacher and FixMatch Self-Training Scheme(s)

anwai98:
- Mean-Teacher training parameters updated
- AdaMT training parameters updated

### DIFF
--- a/experiments/probabilistic_domain_adaptation/livecell/common.py
+++ b/experiments/probabilistic_domain_adaptation/livecell/common.py
@@ -19,14 +19,18 @@ from torch_em.model import UNet2d
 from torch_em.util.prediction import predict_with_padding
 from torchvision import transforms
 from tqdm import tqdm
+from torch_em.util import load_model
 
 CELL_TYPES = ["A172", "BT474", "BV2", "Huh7", "MCF7", "SHSY5Y", "SkBr3", "SKOV3"]
 
 
 #
 # The augmentations we use for the LiveCELL experiments:
-# - weak augmenations: blurring and additive gaussian noise
-# - strong augmentations: TODO
+# - weak augmenations:
+# blurring and additive gaussian noise
+#
+# - strong augmentations:
+# blurring, additive gaussian noise and randon contrast adjustment (FixMatch expects stronger parameters for each)
 #
 
 
@@ -42,9 +46,19 @@ def weak_augmentations(p=0.25):
     return torch_em.transform.raw.get_raw_transform(normalizer=norm, augmentation1=aug)
 
 
-# TODO
-def strong_augmentations():
-    pass
+def strong_augmentations(p=0.5):
+    norm = torch_em.transform.raw.standardize
+    aug1 = transforms.Compose([
+        norm,
+        transforms.RandomApply([torch_em.transform.raw.GaussianBlur(sigma=(0.6, 3.0))], p=p),
+        transforms.RandomApply([torch_em.transform.raw.AdditiveGaussianNoise(
+            scale=(0.05, 0.25), clip_kwargs=False)], p=p/2
+        ),
+        transforms.RandomApply([torch_em.transform.raw.RandomContrast(
+            mean=0.0, alpha=(0.33, 3.0), clip_kwargs=False)], p=p
+        )
+    ])
+    return torch_em.transform.raw.get_raw_transform(normalizer=norm, augmentation1=aug1)
 
 
 #
@@ -55,12 +69,28 @@ def get_unet():
     return UNet2d(in_channels=1, out_channels=1, initial_features=64, final_activation="Sigmoid", depth=4)
 
 
-def load_model(model, ckpt, get_model=get_unet, device=None):
-    model = get_model()
-    model = torch_em.util.get_trainer(ckpt).model
-    if device is not None:
-        model.to(device)
-    return model
+# Computing the Source Distribution for Distribution Alignment
+def compute_class_distribution(root_folder, label_threshold=0.5):
+
+    bg_list, fg_list = [], []
+    total = 0
+
+    files = glob(os.path.join(root_folder, "*"))
+    assert len(files) > 0, f"Did not find predictions @ {root_folder}"
+
+    for pl_path in files:
+        img = imageio.imread(pl_path)
+        img = np.where(img >= label_threshold, 1, 0)
+        _, counts = np.unique(img, return_counts=True)
+        assert len(counts) == 2
+        bg_list.append(counts[0])
+        fg_list.append(counts[1])
+        total += img.size
+
+    bg_frequency = sum(bg_list) / float(total)
+    fg_frequency = sum(fg_list) / float(total)
+    assert np.isclose(bg_frequency + fg_frequency, 1.0)
+    return [bg_frequency, fg_frequency]
 
 
 # use get_model and prediction_function to customize this, e.g. for using it with the PUNet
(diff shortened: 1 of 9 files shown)


experiments/probabilistic_domain_adaptation/livecell/common.py
```diff
@@ -55,9 +55,9 @@ def get_unet():
     return UNet2d(in_channels=1, out_channels=1, initial_features=64, final_activation="Sigmoid", depth=4)
 
 
-def load_model(model, ckpt, state="model_state", device=None):
-    state = torch.load(os.path.join(ckpt, "best.pt"))[state]
-    model.load_state_dict(state)
+def load_model(model, ckpt, get_model=get_unet, device=None):
```
constantinpape, line 58: You can now use this function for loading the model, it takes care of the pytorch 2 things: https://github.com/constantinpape/torch-em/blob/main/torch_em/util/util.py#L204

anwai98: Updated in the recent [commit](https://github.com/constantinpape/torch-em/pull/116/commits/18f289e6d6d8067ed625d80be0107b01bf9ab7ac#diff-414c225623f833e81230091d9429fed78c30d5a5c3f6fd26238ff9587d59202cR126).

torch_em/self_training/fix_match.py
```diff
@@ -0,0 +1,309 @@
...
+        for xu1, xu2 in self.unsupervised_train_loader:
+            xu1, xu2 = xu1.to(self.device), xu2.to(self.device)
+
+            teacher_input, model_input = xu1, xu2
+
+            self.optimizer.zero_grad()
+            # Perform unsupervised training
+            with forward_context():
+                # Compute the pseudo labels.
+                pseudo_labels, label_filter = self.pseudo_labeler(self.model, teacher_input)
```
constantinpape, line 164: We need to be careful here that the pseudo labels are not taken into account when computing gradients. So this should be wrapped into a `with torch.no_grad()` and we should `detach` the `pseudo_labels`. (This might be missing from the mean teacher trainer as well and so should be added there too).

torch_em/self_training/fix_match.py
```diff
@@ -0,0 +1,309 @@
...
+            with forward_context():
+                # We pass the model, the input and the labels to the supervised loss function,
+                # so that how the loss is calculated stays flexible, e.g. to enable ELBO for PUNet.
+                supervised_loss = self.supervised_loss(self.model, xs, ys)
+
+            teacher_input, model_input = xu1, xu2
+            # Perform unsupervised training
+            with forward_context():
+                # Compute the pseudo labels.
+                pseudo_labels, label_filter = self.pseudo_labeler(self.model, teacher_input)
```
constantinpape, line 210: Again, we need to make sure that no gradients are accumulated in the pseudo label computation (see above).

constantinpape (commented):
Thanks! I left a few comments. The most important thing to address is making sure that pseudo labels are not used in gradient computation.

torch_em/self_training/fix_match.py
```diff
@@ -0,0 +1,309 @@
...
+        self.pseudo_labeler = pseudo_labeler
+
+        if source_distribution is None:
+            self.source_distribution = None
+        else:
+            self.source_distribution = torch.FloatTensor(source_distribution).to(self.device)
+
+        self._kwargs = kwargs
+
+    def get_distribution_alignment(self, pseudo_labels, label_threshold=0.5):
```
constantinpape, line 130: I didn't fully go through this, and I assume you adapted it from our previous code for the distro alignment.
Still, it would be good to check this again and to give some comments / formulas here.

constantinpape, line 130: I didn't fully go through this, and I assume you adapted it from our previous code for the distro alignment.
Still, it would be good to check this again and to give some comments / formulas here.

torch_em/self_training/fix_match.py
```diff
@@ -0,0 +1,309 @@
...
+    Parameters:
+        model [nn.Module] -
+        unsupervised_train_loader [torch.DataLoader] -
+        unsupervised_loss [callable] -
+        supervised_train_loader [torch.DataLoader] - (default: None)
+        supervised_loss [callable] - (default: None)
+        unsupervised_loss_and_metric [callable] - (default: None)
+        supervised_loss_and_metric [callable] - (default: None)
+        logger [TorchEmLogger] - (default: SelfTrainingTensorboardLogger)
+        momentum [float] - (default: 0.999)
```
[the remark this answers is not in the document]
anwai98: Added the description (for both FixMatch and MeanTeacher trainers)

torch_em/self_training/fix_match.py
```diff
@@ -0,0 +1,309 @@
...
+        - Returns: loss, metric
+    At least one of unsupervised_loss_and_metric and supervised_loss_and_metric must be given.
+
+    Note: adjust the batch size ratio between the 'unsupervised_train_loader' and 'supervised_train_loader'
+    for setting the ratio between supervised and unsupervised training samples
+
+    Parameters:
+        model [nn.Module] -
+        unsupervised_train_loader [torch.DataLoader] -
+        unsupervised_loss [callable] -
```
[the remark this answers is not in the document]
anwai98: Added the description (for both FixMatch and MeanTeacher trainers)

torch_em/self_training/fix_match.py
anwai98: Fixed it (in both FixMatch and MeanTeacher trainers). Thanks for mentioning!

torch_em/self_training/fix_match.py
anwai98: Added some highlights about distribution alignment from [ReMixMatch](https://arxiv.org/abs/1911.09785) (giving hints about the idea).

anwai98:
Thanks for all the feedback! The expected changes have been made (in both the self-training approaches).

torch_em/self_training/fix_match.py
constantinpape: You need to keep the `forward_context()` here. And I think you can just write
```python
with forward_context(), torch.no_grad():
   ...
```

torch_em/self_training/mean_teacher.py
```diff
@@ -196,11 +197,13 @@ def _train_epoch_unsupervised(self, progress, forward_context, backprop):
 
             teacher_input, model_input = xu1, xu2
 
+            with torch.no_grad():
```
constantinpape, line 200: You should also add the `forward_context()` to the `with` statement. (See also above).

torch_em/self_training/mean_teacher.py
```diff
@@ -244,9 +247,13 @@ def _train_epoch_semisupervised(self, progress, forward_context, backprop):
                 supervised_loss = self.supervised_loss(self.model, xs, ys)
 
             teacher_input, model_input = xu1, xu2
+
+            with torch.no_grad():
```
constantinpape, line 251: `forward_context`, see above.

torch_em/self_training/fix_match.py
```diff
@@ -157,14 +162,20 @@ def _train_epoch_unsupervised(self, progress, forward_context, backprop):
 
             teacher_input, model_input = xu1, xu2
 
+            with torch.no_grad():
```
constantinpape, line 165: We also need the `forward_context` here, see below.

constantinpape (commented):
Looks good overall, but we need to keep the `forward_context` around the pseudo_labeler.

anwai98:
The `forward_context` has been updated for generating the Pseudo Labels. Thanks!

constantinpape (approved):
Thanks, looks all good now!

PR #119, opened 2023-04-06
Add 2D Probabilistic UNet

anwai98:


torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+import numpy as np
+
+import torch
+import torch.nn as nn
+from torch.distributions import Normal, Independent, kl
+
+from torch_em.model import UNet2d
+from torch_em.loss.dice import DiceLossWithLogits
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
constantinpape, line 13: Don't define this globally. Instead pass this as an argument to the functions or arguments that need it.

constantinpape (commented):
Thanks! I just had a brief initial look, and found one obvious thing that needs to be change. I will have a closer look at the rest in the next few days.

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+                input_channels,
+                num_filters,
+                no_convs_per_block,
+                initializers,
+                padding=True,
+                posterior=False,
+                num_classes=None
+            ):
+
+        super(Encoder, self).__init__()
```
constantinpape, line 58: This style of writing the `super` calls is outdated. It's better to just write `super().__init__()` instead.

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+                input_channels,
+                num_filters,
+                no_convs_per_block,
+                latent_dim,
+                initializers,
+                posterior=False,
+                num_classes=None
+            ):
+
+        super(AxisAlignedConvGaussian, self).__init__()
```
constantinpape, line 116: See above.

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+                num_filters,
+                latent_dim,
+                num_output_channels,
+                num_classes,
+                no_convs_fcomb,
+                initializers,
+                use_tile=True
+            ):
+
+        super(Fcomb, self).__init__()
```
constantinpape, line 209: See above.

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+                num_classes=1,
+                num_filters=[32, 64, 128, 192],
+                latent_dim=6,
+                no_convs_fcomb=4,
+                beta=10.0,
+                consensus_masking=False,
+                rl_swap=False
+            ):
+
+        super(ProbabilisticUnet, self).__init__()
```
constantinpape, line 312: See above.

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+
+class Encoder(nn.Module):
+    """
+    A convolutional neural network, consisting of len(num_filters) times a block of no_convs_per_block
+    convolutional layers, after each block a pooling operation is performed.
+    And after each convolutional layer a non-linear (ReLU) activation function is applied.
+    """
+
+    def __init__(
+                self,
```
constantinpape, line 48: Just a minor thing, but the usual style for these inits is to have only one indent:
```
def __init__(
    self,
    aa,
    bb,
    cc,
    dd,
    ...
):
   ...
```
(also relevant in several places below).

torch_em/model/probabilistic_unet.py
```diff
@@ -0,0 +1,445 @@
...
+            z = torch.unsqueeze(z, 3)
+            z = self.tile(z, 3, feature_map.shape[self.spatial_axes[1]])
+
+            # Concatenate the feature map (output of the UNet) and the sample taken from the latent space
+            feature_map = torch.cat((feature_map, z), dim=self.channel_axis)
+            output = self.layers(feature_map)
+            return self.last_layer(output)
+
+
+class ProbabilisticUnet(nn.Module):
```
constantinpape, line 274: Let's call it `ProbabilisticUNet` to be consistent with the other UNet naming.

constantinpape (commented):
I had a bit closer look and left a few more comments. In addition to addressing those we should also add a test, see for example https://github.com/constantinpape/torch-em/blob/main/test/model/test_unet.py.

anwai98:
Thanks for the feedback. The discussions for naming consistency (https://github.com/constantinpape/torch-em/pull/119#discussion_r1161325354), updating the indent (https://github.com/constantinpape/torch-em/pull/119#discussion_r1161325047) and style of `super` calls (https://github.com/constantinpape/torch-em/pull/119#discussion_r1161324677) has been updated.

torch_em/model/probabilistic_unet.py
```diff
@@ -196,17 +194,18 @@ class Fcomb(nn.Module):
...
-        super(Fcomb, self).__init__()
+        self,
+        num_filters,
+        latent_dim,
+        num_output_channels,
+        num_classes,
+        no_convs_fcomb,
+        initializers,
+        use_tile=True,
+        device='cpu'
```
constantinpape, line 205: I would rather set this to `None`, and if it's `None` set it to cpu or gpu depending on `torch.cuda.is_available`.

constantinpape (commented):
One more minor comment, besides this the changes look good.

torch_em/model/probabilistic_unet.py
anwai98: True, this makes sense. Thanks for mentioning. It's fixed now!

constantinpape (approved):
Looks good now!

PR #123, opened 2023-04-27
Add 2D Probabilistic UNet Training Setup

anwai98:
- PUNet Source Training Workflow

> (TODO: evaluation script will follow)

### DIFF
--- a/experiments/probabilistic_domain_adaptation/livecell/common.py
+++ b/experiments/probabilistic_domain_adaptation/livecell/common.py
@@ -23,8 +23,6 @@ from tqdm import tqdm
 from torch_em.util import load_model
 
 CELL_TYPES = ["A172", "BT474", "BV2", "Huh7", "MCF7", "SHSY5Y", "SkBr3", "SKOV3"]
-DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
-
 
 #
 # The augmentations we use for the LiveCELL experiments:
--- a/experiments/probabilistic_domain_adaptation/livecell/common.py
+++ b/experiments/probabilistic_domain_adaptation/livecell/common.py
@@ -86,7 +86,7 @@ def get_unet():
 
 def get_punet():
     return ProbabilisticUNet(input_channels=1, num_classes=1, num_filters=[64, 128, 256, 512],
-                             latent_dim=6, no_convs_fcomb=3, beta=1.0)
+                             latent_dim=6, no_convs_fcomb=3, beta=1.0, rl_swap=True)
 
 
 # Computing the Source Distribution for Distribution Alignment
--- a/experiments/probabilistic_domain_adaptation/livecell/common.py
+++ b/experiments/probabilistic_domain_adaptation/livecell/common.py
@@ -10,6 +10,7 @@ import numpy as np
 import pandas as pd
 import torch
 import torch_em
+from torch_em.model import ProbabilisticUNet
 
 from elf.evaluation import dice_score
 from torch_em.data.datasets.livecell import (get_livecell_loader,
@@ -81,6 +82,11 @@ def strong_augmentations(p=0.5, mode=None):
 
 def get_unet():
     return UNet2d(in_channels=1, out_channels=1, initial_features=64, final_activation="Sigmoid", depth=4)
+
+
+def get_punet():
+    return ProbabilisticUNet(input_channels=1, num_classes=1, num_filters=[64, 128, 256, 512],
+                             latent_dim=6, no_convs_fcomb=3, beta=1.0)
 
 
 # Computing the Source Distribution for Distribution Alignment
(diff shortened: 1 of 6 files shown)


torch_em/self_training/probabilistic_unet_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+            prior_samples=16,
+            supervised_loss=None,
+            supervised_loss_and_metric=None,
+            **kwargs
+    ):
+        super().__init__(**kwargs)
+        assert supervised_loss, supervised_loss_and_metric is not None
+
+        self.supervised_loss = supervised_loss
+        self.supervised_loss_and_metric = supervised_loss_and_metric
```
constantinpape, line 35: just a minor point, but we can just call it `loss` and `loss_and_metric` here, since this trainer is only for supervised training, so we don't need to make a distinction between the supervised and unsupervised cases.

torch_em/self_training/probabilistic_unet_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+    Parameters:
+        clip_posterior [bool] - (default: False)
+        prior_samples [int] - (default: 16)
+        supervised_loss [callable] - (default: None)
+        supervised_loss_and_metric [callable] - (default: None)
+    """
+
+    def __init__(
+            self,
+            clip_posterior=False,
```
constantinpape, line 25: I would suggest to just use a `clipping_value` parameter with default `None`. If set to `None` don't clip, otherwise use the `clipping_value`. (see also comment below).

torch_em/self_training/probabilistic_unet_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+            **kwargs
+    ):
+        super().__init__(**kwargs)
+        assert supervised_loss, supervised_loss_and_metric is not None
+
+        self.supervised_loss = supervised_loss
+        self.supervised_loss_and_metric = supervised_loss_and_metric
+
+        self.clip_posterior = clip_posterior
+        self.clipping_value = 1
```
constantinpape, line 38: `self.clipping_value = clipping_value`

torch_em/self_training/probabilistic_unet_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+
+            with forward_context():
+                # We pass the model, the input and the labels to the supervised loss function, so
+                # that's how the loss is calculated stays flexible, e.g. here to enable ELBO for PUNet.
+                loss = self.supervised_loss(self.model, x, y)
+
+            backprop(loss)
+
+            # To counter the exploding gradients in the posterior net
+            if self.clip_posterior:
```
constantinpape, line 76: `if self.clipping_value is not None`

constantinpape (commented):
I left a few minor comments; looks good overall.

torch_em/self_training/probabilistic_unet_trainer.py
anwai98: Yes, this makes sense. Updated it in the traning scheme as well as the loss. Thanks!

experiments/probabilistic_domain_adaptation/livecell/punet_source.py
```diff
@@ -26,16 +26,15 @@ def _train_cell_type(args, cell_type, device=DEVICE):
         model=model,
         train_loader=train_loader,
         val_loader=val_loader,
-        loss=DummyLoss(),
         metric=DummyLoss(),
```
constantinpape, line 29: It's also better if you do this already in the trainer. Then it's not necessary to do it on the user side. (see comment below)

torch_em/self_training/probabilistic_unet_trainer.py
```diff
@@ -14,28 +14,26 @@ class ProbabilisticUNetTrainer(torch_em.trainer.DefaultTrainer):
...
             prior_samples=16,
-            supervised_loss=None,
-            supervised_loss_and_metric=None,
+            loss=None,
+            loss_and_metric=None,
             **kwargs
     ):
-        super().__init__(**kwargs)
-        assert supervised_loss, supervised_loss_and_metric is not None
+        super().__init__(loss=loss, **kwargs)
```
constantinpape, line 31: set `metric=DummyMetric()` here, see above.

anwai98:
The PUNet Source Workflow is completed now. Just updated the pipeline for source evaluations. Thanks!

experiments/probabilistic_domain_adaptation/livecell/common.py
```diff
@@ -23,6 +23,7 @@
 from torch_em.util import load_model
 
 CELL_TYPES = ["A172", "BT474", "BV2", "Huh7", "MCF7", "SHSY5Y", "SkBr3", "SKOV3"]
+DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
constantinpape, line 26: This is not a good idea to declare globally, and I also don't see that you use it anywhere.

constantinpape (commented):
One minor comment, otherwise this looks good.
I am looking into the test failure here now: https://github.com/constantinpape/torch-em/pull/124

constantinpape (approved):

PR #133, opened 2023-07-11
UNETR Implementation in torch-em

anwai98:
Implementing UNETR from scratch using the following components:
- Encoder from [Segment Anything](https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/modeling/image_encoder.py)
- Decoder from [UNet2d](https://github.com/constantinpape/torch-em/blob/main/torch_em/model/unet.py) in `torch_em`

### DIFF
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -1,0 +1,446 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from typing import Optional, Tuple, Type
+from functools import partial
+
+from segment_anything.modeling.common import LayerNorm2d, MLPBlock
+
+from micro_sam.util import get_sam_model
+
+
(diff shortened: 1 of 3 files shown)


torch_em/model/unetr.py
```diff
@@ -0,0 +1,299 @@
...
+from typing import Tuple
+from functools import partial
+from torch_em.model.unet import Decoder, ConvBlock2d, Upsampler2d
+
+from segment_anything.modeling import ImageEncoderViT
+
+from micro_sam.util import get_sam_model
+
+
+# This class (inherits from Segment Antything) and its supporting functions below lightly adapted from the ViTDet backbone available at: https://github.com/facebookresearch/detectron2/blob/main/detectron2/modeling/backbone/vit.py # noqa
```
constantinpape, line 14: This comment is a bit confusing. Say that you're inheriting from the SAM ViT and what exactly you're changing.

torch_em/model/unetr.py
```diff
@@ -0,0 +1,299 @@
...
+class ViTb_Sam(ImageEncoderViT):
+    def __init__(
```
constantinpape, line 16: Are you changing anything in the init? If not: you do not need to duplicate it.

constantinpape (commented):
Thanks @anwai98. I will do a follow up PR onto this for a couple other things.

PR #146, opened 2023-07-21
Implement UNETR

constantinpape:
Based on the implementation by @anwai98 , with some refactoring by me.


torch_em/model/unetr.py
```diff
@@ -0,0 +1,296 @@
...
+        self.deconv2 = Deconv2DBlock(features_decoder[0], features_decoder[1])
+        self.deconv3 = Deconv2DBlock(features_decoder[1], features_decoder[2])
+
+        self.deconv4 = SingleDeconv2DBlock(features_decoder[-1], features_decoder[-1])
+
+        self.decoder_head = ConvBlock2d(2*features_decoder[-1], features_decoder[-1])
+
+    def preprocess(self, x: torch.Tensor) -> torch.Tensor:
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+        # TODO Why do we have hard-code image normalization here???
```
constantinpape, line 172: @anwai98 Why do you use hard-coded values for the normalization here? In general why is this happening in the network, we usually do this as pre-processing independent of the network.

torch_em/model/unetr.py
```diff
@@ -0,0 +1,296 @@
...
+        raise RuntimeError(
+            "micro_sam is required to initialize the UNETR image encoder from segment anything weights."
+            "Please install it from https://github.com/computational-cell-analytics/micro-sam"
+            "and then rerun your code."
+        )
+    predictor = get_sam_model(model_type=model_type, checkpoint_path=checkpoint_path)
+    _image_encoder = predictor.model.image_encoder
+
+    image_encoder = ViTb_Sam()
+    # FIXME this doesn't work yet because the parameters don't match
```
constantinpape, line 290: @anwai98 I updated the code a bit compared to what you had (I am pretty sure your version wasn't yet working but didn't try it). 
But my current version still isn't working...

anwai98: Ah, okay! The last I checked, the parameters were getting initialized from SAM in the old setup (checked it before and after the function was taking charge to initialize the ViT, the encoder blocks were updating) (however, I think your implementation is smarter and we could keep it to our micro-sam standard)

torch_em/model/unetr.py
constantinpape: I made this optional now.

torch_em/model/unetr.py
constantinpape: This is still not fully working for me, but I will go ahead and merge and we can fix it in a follow up.

PR #148, opened 2023-09-13
Fixes and Updates for UNETR

anwai98:
I checked up for UNETR, and tried to fix the overalls (one final thing is still missing - the switch between ViT models)

To address some previous questions - 
- We do the preprocessing for the inputs to make sure that the dimensions are as SAM's ViT appreciates (1024 * 1024)
- The initialization of SAM works now (the parameters are getting updated)
- Added training script for Cremi (and updated for LiveCell)

In general, UNETR models can now be called from two obvious functions super easily (`from torch_em.model import build_unetr_with_sam_initialization, build_unetr_without_sam_initialization`) (not 100% of the nomenclature though)

### DIFF
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -39,7 +39,7 @@ class ViT_Sam(ImageEncoderViT):
                 "and then rerun your code."
             )
 
-        super().__init__(**kwargs)
+        super().__init__(embed_dim=embed_dim, **kwargs)
         self.chunks_for_projection = global_attn_indexes
         self.in_chans = in_chans
         self.embed_dim = embed_dim
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -21,7 +21,7 @@ except ImportError:
     get_sam_model = None
 
 
-class ViTb_Sam(ImageEncoderViT):
+class ViT_Sam(ImageEncoderViT):
     """Vision Transformer derived from the Segment Anything Codebase (https://arxiv.org/abs/2304.02643):
     https://github.com/facebookresearch/segment-anything/blob/main/segment_anything/modeling/image_encoder.py
     """
@@ -112,10 +112,12 @@ def window_unpartition(
 class UNETR(nn.Module):
     def __init__(
         self,
-        encoder=None,
+        encoder="vit_b",
         decoder=None,
         out_channels=1,
-        use_sam_preprocessing=False,
+        use_sam_preprocessing=True,
+        initialize_from_sam=False,
+        checkpoint_path=None
     ) -> None:
         depth = 3
         initial_features = 64
@@ -127,8 +129,8 @@ class UNETR(nn.Module):
 
         super().__init__()
 
-        if encoder is None:
-            self.encoder = ViTb_Sam(
+        if encoder == "vit_b":
+            self.encoder = ViT_Sam(
                 depth=12,
                 embed_dim=768,
                 img_size=1024,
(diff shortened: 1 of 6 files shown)


anwai98:
And now the final fix is also in (other ViT models can be switched and used as well) ;) Thanks!

torch_em/model/unetr.py
```diff
@@ -112,10 +112,12 @@ def window_unpartition(
...
     def __init__(
         self,
-        encoder=None,
+        encoder="vit_b",
         decoder=None,
         out_channels=1,
-        use_sam_preprocessing=False,
+        use_sam_preprocessing=True,
+        initialize_from_sam=False,
+        checkpoint_path=None
```
constantinpape, line 120: Is there a reason we need both `initialize_from_sam` and `checkpoint_path`? From what I see it would be enough to have `checkpoint_path` (see also comment below).

torch_em/model/unetr.py
```diff
@@ -142,8 +144,51 @@ def __init__(
...
+                qkv_bias=True,
+                use_rel_pos=True,
+                global_attn_indexes=[7, 15, 23, 31],  # type: ignore
+                window_size=14,
+                out_chans=256
+            )
+
         else:
-            self.encoder = encoder
+            raise NotImplementedError
```
constantinpape, line 181: You should throw a `ValueError` with a meaningful message, e.g. `raise ValueError(f"{encoder} is not supported. Currently only vit_b, vit_l, vit_h are supported.")`.

torch_em/model/unetr.py
```diff
@@ -142,8 +144,51 @@ def __init__(
...
+
+        if initialize_from_sam:
```
constantinpape, line 183: I think we don't need `initialize_from_sam`. Isn't it enough to have `if checkpoint_path is not None` here? (And then you don't need to have the assertion.)

torch_em/model/unetr.py
constantinpape: Also, I suggest we call it `encoder_checkpoint_path`.

torch_em/model/unetr.py
```diff
@@ -279,21 +324,12 @@ def forward(self, x):
...
-def build_unetr_with_sam_intialization(out_channels=1, model_type="vit_b", checkpoint_path=None):
-    if get_sam_model is None:
-        raise RuntimeError(
-            "micro_sam is required to initialize the UNETR image encoder from segment anything weights."
-            "Please install it from https://github.com/computational-cell-analytics/micro-sam"
-            "and then rerun your code."
-        )
-    predictor = get_sam_model(model_type=model_type, checkpoint_path=checkpoint_path)
-    _image_encoder = predictor.model.image_encoder
+def build_unetr_with_sam_initialization(out_channels=1, model_type="vit_b", checkpoint_path=None):
```
constantinpape, line 327: We can get rid of these functions. The current design is simple enough and we don't need to wrap it in a function.

constantinpape (commented):
Looks good overall! Just a few minor comments.

torch_em/model/unetr.py
anwai98: > Is there a reason we need both initialize_from_sam and checkpoint_path? From what I see it would be enough to have checkpoint_path (see also comment below).

Yes, this makes total sense. Fixed it under one argument - `encoder_checkpoint_path`

torch_em/model/unetr.py
anwai98: Okie. In that case, I will document it in the directory the correct usage for different ViT models with and without initialization.

constantinpape:
The test failure is unrelated. I will merge and fix that in a separate PR.

PR #153, opened 2023-10-05
Add PanNuke Dataloader

anwai98:
- Adding the dataloader for the [PanNuke](https://warwick.ac.uk/fac/cross_fac/tia/data/pannuke
) dataset (histopathology domain)

### DIFF
--- a/torch_em/data/datasets/pannuke.py
+++ b/torch_em/data/datasets/pannuke.py
@@ -40,18 +40,19 @@ def _download_pannuke_dataset(path, download, folds):
         print(f"Unzipping the PanNuke dataset in {tmp_fold} directories...")
         util.unzip(os.path.join(path, f"{tmp_fold}.zip"), os.path.join(path, f"{tmp_fold}"), True)
 
+        _convert_to_hdf5(path, tmp_fold)
 
-def _convert_to_hdf5(path, folds):
-    for tmp_fold in folds:
-        if os.path.exists(os.path.join(path, f"pannuke_{tmp_fold}.h5")):
-            return
 
-        print(f"Converting the {tmp_fold} into h5 file format...")
-        tmp_name = tmp_fold.split("_")[0] + tmp_fold.split("_")[1]  # name of a particular sub-directory (per fold)
-        img_path = glob(os.path.join(path, tmp_fold, "*", "images", tmp_name, "images.npy"))[0]
-        gt_path = glob(os.path.join(path, tmp_fold, "*", "masks", tmp_name, "masks.npy"))[0]
+def _convert_to_hdf5(path, fold):
+    if os.path.exists(os.path.join(path, f"pannuke_{fold}.h5")) is True:
+        return
 
-        with h5py.File(os.path.join(path, f"pannuke_{tmp_fold}.h5"), "w") as f:
+    print(f"Converting {fold} into h5 file format...")
+    img_paths = glob(os.path.join(path, "**", "images.npy"), recursive=True)
+    gt_paths = glob(os.path.join(path, "**", "masks.npy"), recursive=True)
+
+    for img_path, gt_path in zip(img_paths, gt_paths):
+        with h5py.File(os.path.join(path, f"pannuke_{fold}.h5"), "w") as f:
             f.create_dataset("images", data=np.load(img_path).transpose(3, 0, 1, 2))
             f.create_dataset("masks", data=np.load(gt_path).transpose(3, 0, 1, 2))
 
@@ -74,7 +75,6 @@ def get_pannuke_dataset(
         assert isinstance(rois, dict)
 
     _download_pannuke_dataset(path, download, folds)
-    _convert_to_hdf5(path, folds)
 
     data_paths = [os.path.join(path, f"pannuke_{f_idx}.h5") for f_idx in folds]
     data_rois = [rois.get(f_idx, np.s_[:, :, :]) for f_idx in folds]
--- a/torch_em/data/datasets/pannuke.py
+++ b/torch_em/data/datasets/pannuke.py
@@ -19,7 +19,11 @@ URLS = {
 
 
 # TODO
-CHECKSUM = None
+CHECKSUM = {
+    "fold_1": "6e19ad380300e8ce9480f9ab6a14cc91fa4b6a511609b40e3d70bdf9c881ed0b",
+    "fold_2": "5bc540cc509f64b5f5a274d6e5a245527dbd3e6d3155d43555115c5d54709b07",
+    "fold_3": "c14d372981c42f611ebc80afad01702b89cad8c1b3089daa31931cf5a4b1a39d"
+}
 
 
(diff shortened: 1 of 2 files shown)


anwai98:
At the current stage, it looks like it's working. It would be great to have an overlook on this.
In brief (the current stage): It does the download and all the formattings end-to-end, to just expect the `path` where the data is stored (or supposed to be), and with `download=True` does all the rest stuff.

There's a rather complicated label transformation taking care of a lot of stuff (briefed in the respective function in the file itself). Let me know how it looks.

torch_em/data/datasets/pannuke.py
```diff
@@ -0,0 +1,149 @@
...
+def _convert_to_hdf5(path, folds):
+    for tmp_fold in folds:
+        if os.path.exists(os.path.join(path, f"pannuke_{tmp_fold}.h5")):
+            return
+
+        print(f"Converting the {tmp_fold} into h5 file format...")
+        tmp_name = tmp_fold.split("_")[0] + tmp_fold.split("_")[1]  # name of a particular sub-directory (per fold)
+        with h5py.File(os.path.join(path, f"pannuke_{tmp_fold}.h5"), "w") as f:
+            img_path = glob(os.path.join(path, tmp_fold, "*", "images", tmp_name, "images.npy"))[0]
+            gt_path = glob(os.path.join(path, tmp_fold, "*", "masks", tmp_name, "masks.npy"))[0]
```
constantinpape, line 49: You only take a single of the files here? Is this on purpose?
I would have assumed that the code should look something like this instead:
```python
img_paths = sorted(glob(..., "images.npy))  # imported to do sorted here, so that you get the same order for masks and images
gt_paths = sorted(glob(..., "masks.npy))  # imported to do sorted here, so that you get the same order for masks and images
for i, (img_path, gt_path) in enumerate(zip(img_paths, gt_paths)):
  img = np.load(img_path)
  gt = np.load(gt_path)
  assert img,shape == gt.shape  # or similar; but make sure that the shapes match, you might need to account for a different number of channels
  out_path = os.path.join(..., f"pannuke_{tmp_fold}_{i}.h5"
  with h5py.File(out_path, "w") as f:
      ...
```
That way you would use all the images in a given fold

torch_em/data/datasets/pannuke.py
```diff
@@ -0,0 +1,149 @@
...
+
+            f.create_dataset("images", data=np.load(img_path).transpose(3, 0, 1, 2))
+            f.create_dataset("masks", data=np.load(gt_path).transpose(3, 0, 1, 2))
```
constantinpape, line 52: I am a bit confused by how many channels we have here. I would have assumed that the image data has 3 dimensions (2 spatial ones and 1 channel dimension) and that the segmentation has 2 dimensions (only the spatial ones).

constantinpape: Ah never mind about the comment on the segmentations / labels, I saw the label trafo now. Still, I would only expect 3 dimensions, not four.

constantinpape, line 52: Use `compression="gzip"` when storing hdf5 datasets. This will save a lot of memory, especially for the segmentation data.

torch_em/data/datasets/pannuke.py
```diff
@@ -0,0 +1,149 @@
...
+        path=path,
+        patch_shape=patch_shape,
+        folds=folds,
+        rois=rois,
+        download=download,
+        **dataset_kwargs)
+    return torch_em.get_data_loader(ds, batch_size=batch_size, **loader_kwargs)
+
+
+def label_trafo(labels):
```
constantinpape, line 110: I don't think this gives you a correct instance segmentation. Even if it does, it's too complex, you don't need the `np.where`. I would write it like this:

```python
segmentation = np.zeros(labels.shape[1:])
max_id = 0
for label_channel in labels[:-1]:  # from what I understand we can just ignore the last channel because it encodes background
    this_labels = vigra.analysis.labelImage(label_channel)  # connected components to make sure we have an instance segmentation
    foreground = this_labels > 0
    segmentation[foreground] = this_labels[foreground] + max_id
    max_id = segmentation.max()
```

constantinpape (commented):
There are a couple of things that don't look correct to me, I left comments about that.
Another general note: I think it would make more sense to apply the transformation that make instance and semantic labels out of the masks when creating the data; this makes it easier to later use different kinds of label transforms on the data (e.g. connected components for sub-patches).

torch_em/data/datasets/pannuke.py
anwai98: Ah, it's to provide the option to (only) use specific folds of the dataset (there are 3 in total now) (I took the inspiration for this from cremi). Do you think we should do the download and h5 conversion at once for all the folds already?

torch_em/data/datasets/pannuke.py
anwai98: Just to be sure, the input image and input label dimensions look like this (S x H x W x C)

where,
- S is the number of slices
- C is the number of channels
    - for the input images - it's RGB (3)
    - for the input labels - it's 6 (5 tissue types and last channel is the background)

torch_em/data/datasets/pannuke.py
constantinpape: > Ah, it's to provide the option to (only) use specific folds of the dataset (there are 3 in total now) (I took the inspiration for this from cremi).

That's not what I mean. Having the separate folds is ok. But with the `glob` here and then accessing only the first it looks like you're just selecting one image.
```python
glob(os.path.join(path, tmp_fold, "*", "images", tmp_name, "images.npy"))[0]
```
Maybe I am also understanding the data organization wrong and it's many images stacked. But if that is the case we should save them in separate files to match the data organization expected by torch_em better.

torch_em/data/datasets/pannuke.py
anwai98: I refactored the snippets a slight bit with some minor updates (the major fix is on `np.where`)

torch_em/data/datasets/pannuke.py
```diff
@@ -0,0 +1,149 @@
...
+    if os.path.exists(os.path.join(path, f"pannuke_{fold}.h5")) is True:
+        return
+
+    print(f"Converting {fold} into h5 file format...")
+    img_paths = glob(os.path.join(path, "**", "images.npy"), recursive=True)
+    gt_paths = glob(os.path.join(path, "**", "masks.npy"), recursive=True)
+
+    for img_path, gt_path in zip(img_paths, gt_paths):
+        with h5py.File(os.path.join(path, f"pannuke_{fold}.h5"), "w") as f:
+            f.create_dataset("images", data=np.load(img_path).transpose(3, 0, 1, 2))
```
anwai98, line 56:
```python
# chunks: (3. 1, 256, 256) - C x 1 x H x W
chunks = (data.shape[-1], 1) + data.shape[:2]
f.create_dataset(..., compression="gzip", chunks=chunks)
```

PR #155, opened 2023-10-16
Add Monai's UNETR

anwai98:
- Benchmarking from monai's unetr implementation
- In `experiments/vision-transformer/unetr/results`, the row headers are the source models, and the column headers are for the cell-type-wise inference

> PS: models from `torch-em` and `monai` (both training from scratch) have been tried for 10k iterations for two cell types


anwai98:
We finally have the benchmarking [results](https://docs.google.com/spreadsheets/d/1OTrEf5YKuAigpJ1bT9b3593xm1ydrZCrpEUro3ab3IU/edit#gid=0). The script now looks good to go (the negative values occured how, I am not sure - but they were with a pretty shallow training - for 1k iterations previously). The current results are with 100k iterations (more details of the training heuristic in the sheet linked above)

constantinpape (approved):

PR #156, opened 2023-10-17
Add BCSS Dataset

anwai98:
(WIP) Implementation for the Breast Cancer Semantic Segmentation dataset.
(TODOs):
- ~need to fix data download using `gdown` (download limit for folders is upto 50 files, need to take care of that)~ warning raised
- [x] need to take care of the semantic instances (there are more than one tissue region segmentations available, need to add an optional argument to request for all instances / some instances, as desired)

### DIFF
--- a/scripts/datasets/check_bcss.py
+++ b/scripts/datasets/check_bcss.py
@@ -1,3 +1,8 @@
+import numpy as np
+from typing import Optional, List
+
+import vigra
+
 from torch_em.util.debug import check_loader
 from torch_em.data.datasets import get_bcss_loader
 
@@ -6,17 +11,40 @@ from torch_em.data.datasets import get_bcss_loader
 BCSS_ROOT = "/scratch/projects/nim00007/data/bcss/"
 
 
+class BCSSLabelTrafo():
+    def __init__(
+            self,
+            label_choices: Optional[List[int]] = None
+    ):
+        self.label_choices = label_choices
+
+    def __call__(
+            self,
+            labels: np.ndarray
+    ) -> np.ndarray:
+        """Returns the transformed labels (use-case for SAM)
+        """
+        if self.label_choices is not None:
+            labels[~np.isin(labels, self.label_choices)] = 0
+            segmentation, _, _ = vigra.analysis.relabelConsecutive(labels.astype("uint64"))
+        else:
+            segmentation, _, _ = vigra.analysis.relabelConsecutive(labels)
+
+        return segmentation
+
+
 # NOTE: the bcss data cannot be downloaded automatically.
 # you need to download it yourself from https://bcsegmentation.grand-challenge.org/BCSS/
 def check_bcss():
-    loader = get_bcss_loader(
+    # loader below checks for selective labels
+    chosen_label_loader = get_bcss_loader(
         path=BCSS_ROOT,
         patch_shape=(512, 512),
         batch_size=2,
-        label_choices=None,
-        download=False
+        download=False,
+        label_transform=BCSSLabelTrafo(label_choices=[0, 1, 2])
     )
-    check_loader(loader, 8, instance_labels=True, rgb=True)
+    check_loader(chosen_label_loader, 8, instance_labels=True, rgb=True, plt=True, save_path="./bcss.png")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


torch_em/data/datasets/bcss.py
```diff
@@ -0,0 +1,128 @@
...
+    elif os.path.exists(os.path.join(path, "0_Public-data-Amgad2019_0.25MPP", "rgbs_colorNormalized")):
+        image_paths = sorted(glob(os.path.join(path, "0_Public-data-Amgad2019_0.25MPP", "rgbs_colorNormalized", "*")))
+        label_paths = sorted(glob(os.path.join(path, "0_Public-data-Amgad2019_0.25MPP", "masks", "*")))
+    else:
+        raise ValueError(
+            "Please check the image directory. If downloaded from gdrive, it's named \"rgbs_colorNormalized\", if from github it's named \"images\""
+        )
+    assert len(image_paths) == len(label_paths)
+
+    label_trafo = BCSSLabelTrafo(label_choices=label_choices)
```
constantinpape, line 88: This will lead to problems if the user gives a label trafo.

torch_em/data/datasets/bcss.py
```diff
@@ -0,0 +1,128 @@
...
+    ) -> np.ndarray:
+        """Returns the transformed labels
+        """
+        if self.label_choices is not None:
+            _segmentation = np.zeros(labels.shape)
+            for uq_label in self.label_choices:
+                _segmentation[labels == uq_label] = labels[labels == uq_label]
+            segmentation, _, _ = vigra.analysis.relabelConsecutive(_segmentation.astype("uint64"))
+        else:
+            segmentation, _, _ = vigra.analysis.relabelConsecutive(labels)
```
constantinpape, line 47: If these are semantic labels then we shouldn't do a relabel!
We should discuss tomorrow.

torch_em/data/datasets/bcss.py
```diff
@@ -0,0 +1,128 @@
...
+    def __call__(
+            self,
+            labels: np.ndarray
+    ) -> np.ndarray:
+        """Returns the transformed labels
+        """
+        if self.label_choices is not None:
+            _segmentation = np.zeros(labels.shape)
+            for uq_label in self.label_choices:
+                _segmentation[labels == uq_label] = labels[labels == uq_label]
```
constantinpape, line 44: The three lines above are a very complex way of writing
```
labels[~np.isin(labels, self.label_choices)] = 0
```
(to iterate the point again: avoid loops if you can and go for built-in numpy (or torch) functionality instead, which is both faster and easier to read)

anwai98: Okay, noted. Thanks! (fixed it with a recent commit)

torch_em/data/datasets/bcss.py
anwai98: Okay. Let's discuss this tomorrow what's the best strategy to take care of this.

torch_em/data/datasets/bcss.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+URL = "https://drive.google.com/drive/folders/1zqbdkQF8i5cEmZOGmbdQm-EP8dRYtvss?usp=sharing"
+
+
+# TODO
+CHECKSUM = None
+
+
+class BCSSLabelTrafo():
```
anwai98, line 22: Write label tissue types in the dataset docstring, do the labe trafo in SAM

scripts/datasets/check_bcss.py
```diff
@@ -0,0 +1,51 @@
...
+            self,
+            label_choices: Optional[List[int]] = None
+    ):
+        self.label_choices = label_choices
+
+    def __call__(
+            self,
+            labels: np.ndarray
+    ) -> np.ndarray:
+        """Returns the transformed labels (use-case for SAM)
```
constantinpape, line 25: For the same use-case we should use connected components (`skimage.measure.label`) instead of just relabeling.
This way we can separate disconnected pieces, and have an instance segmentation, which will be better for SAM.
(Since this is just the sample script it's not important to update this here, but please keep it in mind for SAM.)

constantinpape (approved):

PR #157, opened 2023-10-18
Updates to MoNuSeg

anwai98:
- Added processing for the monuseg inputs (including xml to numpy array conversion) (kept the expected directory formats consistent)
- (WIP) Add the test split
- (TODO) Check the organ splits from the challenge's metadata and make those splits (mentioned in TODOs)

### DIFF
--- a/torch_em/data/datasets/monuseg.py
+++ b/torch_em/data/datasets/monuseg.py
@@ -1,12 +1,9 @@
 import os
 import shutil
-import numpy as np
 from tqdm import tqdm
 from glob import glob
-from xml.dom import minidom
 
 import imageio.v2 as imageio
-from skimage.draw import polygon
 
 import torch_em
 from torch_em.data.datasets import util
@@ -40,48 +37,6 @@ def _download_monuseg(path, download, split):
     _process_monuseg(path, split)
 
 
-def generate_labeled_array(shape, xml_file):
-    """Function taken from: https://github.com/rshwndsz/hover-net/blob/master/lightning_hovernet.ipynb
-
-    Given image shape and path to annotations (xml file), generatebit mask with the region inside a contour being white
-        shape: The image shape on which bit mask will be made
-        xml_file: path relative to the current working directory where the xml file is present
-
-    Returns:
-        An image of given shape with region inside contour being white..
-    """
-    # DOM object created by the minidom parser
-    xDoc = minidom.parse(xml_file)
-
-    # List of all Region tags
-    regions = xDoc.getElementsByTagName('Region')
-
-    # List which will store the vertices for each region
-    xy = []
-    for region in regions:
-        # Loading all the vertices in the region
-        vertices = region.getElementsByTagName('Vertex')
-
-        # The vertices of a region will be stored in a array
-        vw = np.zeros((len(vertices), 2))
-
-        for index, vertex in enumerate(vertices):
-            # Storing the values of x and y coordinate after conversion
-            vw[index][0] = float(vertex.getAttribute('X'))
-            vw[index][1] = float(vertex.getAttribute('Y'))
-
-        # Append the vertices of a region
-        xy.append(np.int32(vw))
-
-    # Creating a completely black image
-    mask = np.zeros(shape, np.float32)
-
-    for i, contour in enumerate(xy):
-        r, c = polygon(np.array(contour)[:, 1], np.array(contour)[:, 0], shape=shape)
-        mask[r, c] = i
-    return mask
-
-
 def _process_monuseg(path, split):
     util.unzip(os.path.join(path, f"monuseg_{split}.zip"), path)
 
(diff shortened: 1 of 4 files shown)


torch_em/data/datasets/monuseg.py
```diff
@@ -1,39 +1,89 @@
 import os
+import shutil
+from tqdm import tqdm
+from glob import glob
+
+import imageio.v2 as imageio
```
constantinpape, line 6: We should switch to `imageio.v3`.

torch_em/data/datasets/monuseg.py
```diff
@@ -45,14 +95,35 @@ def get_monuseg_dataset(
...
     dataset = get_monuseg_dataset(
-        path, patch_shape, download=download,
+        path, patch_shape, split, download=download,
         offsets=offsets, boundaries=boundaries, binary=binary, **ds_kwargs
     )
     loader = torch_em.get_data_loader(dataset, batch_size, **loader_kwargs)
     return loader
+
+
+def main():
```
constantinpape, line 111: This is not a good practice. Libraries should not contain scripts inside. Instead add a small script to test it here: https://github.com/constantinpape/torch-em/tree/main/scripts/datasets

constantinpape (commented):
Hi @anwai98 ,
not sure if this is ready to review yet, but I had a quick look. In addition to the comments in the code, I tried the download, but it fails with
```
Access denied with the following error:

        Cannot retrieve the public link of the file. You may need to change
        the permission to 'Anyone with the link', or have had many accesses. 

You may still be able to access the file from the browser:

         https://drive.google.com/uc?export=download&id=1ZgqFJomqQGNnsx7w7QBzQQMVA16lbVCA 
```

torch_em/data/datasets/monuseg.py
anwai98: Thanks for the heads up. Would check all the dataloaders in scripts from now.

anwai98:
> Hi @anwai98 , not sure if this is ready to review yet, but I had a quick look. In addition to the comments in the code, I tried the download, but it fails with
> 
> ```
> Access denied with the following error:
> 
>         Cannot retrieve the public link of the file. You may need to change
>         the permission to 'Anyone with the link', or have had many accesses. 
> 
> You may still be able to access the file from the browser:
> 
>          https://drive.google.com/uc?export=download&id=1ZgqFJomqQGNnsx7w7QBzQQMVA16lbVCA 
> ```

Hmm, pretty interesting. I just tried it again and the download works for me. Is it related to the `gdown` version?
It works with the following version for me:
```
gdown                     4.7.1              pyhd8ed1ab_0    conda-forge
```

constantinpape (commented):
This looks good in principle, but do you still want to implement the organ selection before we merge it?

anwai98:
> This looks good in principle, but do you still want to implement the organ selection before we merge it?

Yes, I wanted to complete the setup, hence took it back as a draft. Would take care of it and put it back to review later. Thanks.

torch_em/data/datasets/monuseg.py
```diff
@@ -1,57 +1,134 @@
...
 ):
-    _download_monuseg(path, download)
+    """Dataset from https://monuseg.grand-challenge.org/Data/
+    """
+    _download_monuseg(path, download, split)
+
+    image_paths = sorted(glob(os.path.join(path, "images", split, "*")))
+    label_paths = sorted(glob(os.path.join(path, "labels", split, "*")))
+
+    if split == "train" and organ_type is not None:
```
constantinpape, line 104: What about the test split? Don't we have the organ_types for it? If not, we should throw an error if the split is test and an organ_type is requested.

torch_em/data/datasets/monuseg.py
```diff
@@ -1,57 +1,134 @@
...
+        # get all patients for multiple organ selection
+        all_organ_splits = sum([ORGAN_SPLITS[o] for o in organ_type], [])
 
-    image_path = os.path.join(path, "images")
-    label_path = os.path.join(path, "labels")
+        image_paths = [
+            _path for _path in image_paths if os.path.split(_path)[-1].split(".")[0] in all_organ_splits
```
constantinpape, line 109: instead of
```python
os.path.split(_path)[-1].split(".")[0]
```
you can do
```python
Path(_path).stem
```
after doing
```python
from pathlib import Path
```
at the top. That achieves the same, but is much more readable.

torch_em/data/datasets/monuseg.py
anwai98: We don't have a given organ split for MoNuSeg. I will raise an error for the test split. Thanks.

torch_em/data/datasets/monuseg.py
anwai98: Ahha that's cool. Thanks for sharing!

constantinpape (approved):

PR #158, opened 2023-10-18
Add MoNuSAC Dataset

anwai98:
(WIP) Haven't been tested, as the `generate_labeled_array` depends on the MoNuSeg PR (will test and put to review once this works)


anwai98:
There's an error thrown by an input image (`TCGA-5P-A9K0-01Z-00-DX1_3.tif`) of shape `(185, 497, 4)`, while the dataloader requests for a patch shape of (512, 512), stating:
```
  File "/home/nimanwai/torch-em/scripts/datasets/check_monusac.py", line 32, in <module>
    check_monusac()
  File "/home/nimanwai/torch-em/scripts/datasets/check_monusac.py", line 18, in check_monusac
    check_loader(train_loader, 8, instance_labels=True, rgb=True, plt=True, save_path="./monusac_train.png")
  File "/home/nimanwai/torch-em/torch_em/util/debug.py", line 127, in check_loader
    _check_plt(loader, n_samples, instance_labels, save_path=save_path)
  File "/home/nimanwai/torch-em/torch_em/util/debug.py", line 17, in _check_plt
    for ii, (x, y) in enumerate(loader):
  File "/scratch/usr/nimanwai/mambaforge/envs/sam/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 634, in __next__
    data = self._next_data()
  File "/scratch/usr/nimanwai/mambaforge/envs/sam/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 678, in _next_data
    data = self._dataset_fetcher.fetch(index)  # may raise StopIteration
  File "/scratch/usr/nimanwai/mambaforge/envs/sam/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 51, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/scratch/usr/nimanwai/mambaforge/envs/sam/lib/python3.10/site-packages/torch/utils/data/_utils/fetch.py", line 51, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/home/nimanwai/torch-em/torch_em/data/image_collection_dataset.py", line 167, in __getitem__
    raw, labels = self._get_sample(index)
  File "/home/nimanwai/torch-em/torch_em/data/image_collection_dataset.py", line 134, in _get_sample
    raw, label = self._ensure_patch_shape(raw, label, have_raw_channels, have_label_channels, channel_first)
  File "/home/nimanwai/torch-em/torch_em/data/image_collection_dataset.py", line 106, in _ensure_patch_shape
    raise NotImplementedError("Padding is not implemented for data with channels")
NotImplementedError: Padding is not implemented for data with channels
```
whereas, if I request for a patch shape of (128. 128), it works just fine.

What do you think about this @constantinpape? (I think the issue is that the paddding implementation for multi-channel inputs is missing)

anwai98:
@constantinpape The padding for images with raw and label channels has been taken care of now. Let me know how this looks now.

torch_em/data/raw_image_collection_dataset.py
```diff
@@ -81,11 +81,9 @@ def _sample_bounding_box(self, shape):
     def _ensure_patch_shape(self, raw, have_raw_channels):
         shape = raw.shape
         if any(sh < psh for sh, psh in zip(shape, self.patch_shape)):
```
constantinpape, line 83: We need to also do this here: (see the corresponding function in `image_collection_dataset`):
```
if have_raw_channels and channel_first:
    shape = shape[1:]
```
and hence we also need to pass `channel_first` as an argument to this function.

torch_em/data/raw_image_collection_dataset.py
```diff
@@ -81,11 +81,9 @@ def _sample_bounding_box(self, shape):
...
-            if have_raw_channels:
-                raise NotImplementedError("Padding is not implemented for data with channels")
-            assert len(shape) == len(self.patch_shape)
             pw = [(0, max(0, psh - sh)) for sh, psh in zip(shape, self.patch_shape)]
-            raw = np.pad(raw, pw)
+            pw_raw = [*pw, (0, 0)] if have_raw_channels else pw
```
constantinpape, line 85: Also need to adapt this for `channel_first`, see above.

constantinpape (commented):
Besides the comments I left in the code: this data is saved as RGBA. That doesn't make sense, the alpha channel (A) does not contain any information and we should get rid of it.

torch_em/data/image_collection_dataset.py
```diff
@@ -102,11 +102,10 @@ def _ensure_patch_shape(self, raw, labels, have_raw_channels, have_label_channel
         if have_raw_channels and channel_first:
             shape = shape[1:]
         if any(sh < psh for sh, psh in zip(shape, self.patch_shape)):
-            if have_raw_channels or have_label_channels:
-                raise NotImplementedError("Padding is not implemented for data with channels")
-            assert len(shape) == len(self.patch_shape)
             pw = [(0, max(0, psh - sh)) for sh, psh in zip(shape, self.patch_shape)]
-            raw, labels = np.pad(raw, pw), np.pad(labels, pw)
+            pw_raw = [*pw, (0, 0)] if have_raw_channels else pw
```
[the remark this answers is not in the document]
anwai98: Okay. I updated it in the latest commit. I also see that we currently don't support multiple channels for labels, hence would add padding for it as a TODO for now.

anwai98:
> Besides the comments I left in the code: this data is saved as RGBA. That doesn't make sense, the alpha channel (A) does not contain any information and we should get rid of it.

Okay. So for this, what would you recommend? (to manually remove the alpha channels from the tif images? ~ using a functionality)

torch_em/data/datasets/monusac.py
```diff
@@ -0,0 +1,183 @@
...
+
+def _check_channel_consistency(path, split):
+    "The provided tif images have RGBA channels, check and remove the alpha channel"
+    all_image_path = glob(os.path.join(path, "images", split, "*.tif"))
+    for image_path in all_image_path:
+        image = imageio.imread(image_path)
+
+        # the provided tif images are provided as channels last.
+        if image.shape[-1] == 4:
+            rgb_image = rgba2rgb(image)
```
constantinpape, line 76: We shouldn't use the `skimage.color` functionality here. This is meant for photos, where having RGBA images actually has a meaning. But we deal with scientific images, where RGB / RGBA does not make sense. The reason we have RBGA at all in this data is a bug (by the data producers)...

In addition to this, your code could be a bit better; you're currently not handling the case where `image.shape[-1] != 4`, in that case everything would fail. (I assume that doesn't happen because all data has the same format, but still it's better to write the code more robust).

Here's how I would write it, including just getting rid of the last channel to remove the alpha channel:
```python
image = imageio.imread(image_path)
assert image.shape[-1] == 4, f"Image has an unexpected shape: {image.shape}; expected 4 channels."
image = image[..., :-1]  # get rid of the alpha channel
imageio.imwrite(image_path, image)
```

constantinpape (commented):
Looks good now, you should just change how the RGBA data is converted.

torch_em/data/datasets/monusac.py
constantinpape: And just for reference, you can see the implementation of `RGBA2RGB`[here](https://github.com/scikit-image/scikit-image/blob/v0.22.0/skimage/color/colorconv.py#L171-L239). It uses alpha blending to get rid of the alpha channel -> we don't need that and it may result in wrong results for us.

anwai98: Ahha okay, that's good to know. Would take care with the conversions. Updated the current functionality as recommended. Thanks!

constantinpape (approved):

PR #164, opened 2023-10-25
Add splits for BCSS dataset

anwai98:
- Making train / val split using random splitting on the fly, and test data is pre-defined in a global list, which has exclusive 10% (of the entire data) patient slides.

### DIFF
--- a/torch_em/data/datasets/bcss.py
+++ b/torch_em/data/datasets/bcss.py
@@ -3,8 +3,9 @@ import shutil
 from glob import glob
 from pathlib import Path
 
-import torch
+from sklearn.model_selection import train_test_split
 
+import torch
 import torch_em
 from torch_em.data.datasets import util
 from torch_em.data import ImageCollectionDataset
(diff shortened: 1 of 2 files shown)


torch_em/data/datasets/bcss.py
```diff
@@ -82,12 +146,12 @@ def get_bcss_dataset(path, patch_shape, download=False, label_dtype=torch.int64,
 
 
 def get_bcss_loader(
-        path, patch_shape, batch_size, download=False, label_dtype=torch.int64, **kwargs
+        path, patch_shape, batch_size, split, val_fraction=0.2, download=False, label_dtype=torch.int64, **kwargs
```
constantinpape, line 149: Would it be a lot of effort for you to make `split` optional, and if it's not passed return all data?

constantinpape (approved):

torch_em/data/datasets/bcss.py
anwai98: Done ;)

constantinpape (approved):

PR #166, opened 2023-11-02
UNETR Training with Affinities

anwai98:
Add unetr training script with affinities for livecell dataset

> TODO: would be merging to `dev` instead of `main` in `constantinpape/torch-em`


anwai98:
@constantinpape this looks good to go!

PR #170, opened 2023-11-27
Add Raw Transform - nnUNet

anwai98:
Creating a base class and child class (for CT/PET-CT normalization scheme) for the raw transformation implementation.
(maybe it's better if I handle inputs as tensors in the raw transform (instead of arrays as of now), but not 100% sure)


anwai98:
@constantinpape This looks good to go (as discussed)

constantinpape (approved):

constantinpape (commented):
I realized there is a bit more subtlety to this.

anwai98:
I updated the raw transform now. It expects channels (modalities) first, and does the transformation automatically for the entire multi-modal input(s) (considering the inputs are 2 * H * W * D - for our case). Is this desired?

anwai98:
This is taken care of now (with the mentions). Thanks!

PR #177, opened 2023-12-04
Distance label trafos

constantinpape:
Fyi @anwai98 this implements the label transformation for per object distances. I have changed a few names compared to the previous version of this to keep it in line with the other transform. (But it should be easy to adapt the training code to this once you're going back to the instance segmentation.)


anwai98:
Thanks @constantinpape. Once you think is good to go as well, let's merge it to `main`. (I will take care of the adpatation in the respective PRs)

PR #178, opened 2023-12-06
Add distance based loss

constantinpape:


torch_em/loss/distance_based.py
```diff
@@ -0,0 +1,63 @@
...
+    def forward(self, input_, target):
+        assert input_.shape == target.shape, input_.shape
+        assert input_.shape[1] == 3, input_.shape
+
+        # IMPORTANT: preserve the channels!
+        # Otherwise the Dice Loss will do all kinds of shennanigans.
+        # Because it always interprets the first axis as channel,
+        # and treats it differently (sums over it independently).
+        # This will lead to a very large dice loss that dominates over everything else.
+        fg_input, fg_target = input_[:, 0:1], target[:, 0:1]
```
constantinpape, line 38: This is crucial, if we don't keep the channels then the dice loss does wrong things.

torch_em/loss/distance_based.py
```diff
@@ -0,0 +1,63 @@
...
+            mask = fg_target
+            bdist_loss = self.distance_loss(bdist_input * mask, bdist_target * mask)
+        else:
+            bdist_loss = self.distance_loss(bdist_input, bdist_target)
+
+        overall_loss = fg_loss + cdist_loss + bdist_loss
+        return overall_loss
+
+
+class DiceBasedDistanceLoss(DistanceLoss):
```
constantinpape, line 59: I by accident tried this (using dice loss both for the foreground and distance loss term) and this seems to work really good. We should give it a try.

torch_em/model/unetr.py
```diff
@@ -27,7 +27,8 @@ def __init__(
         out_channels=1,
         use_sam_stats=False,
         use_mae_stats=False,
-        encoder_checkpoint_path=None
+        encoder_checkpoint_path=None,
+        final_activation=None,
```
constantinpape, line 31: As discussed, it's not a good idea to hard-code the final activation to sigmoid. I updated this to use the same logic and syntax as in our U-Net implementation.

constantinpape (commented):
cc @anwai98

torch_em/loss/distance_based.py
anwai98: Okay, noted!

PR #179, opened 2023-12-06
Fix UNETR Training - Update Rescaling Inputs

anwai98:
This PR takes care of avoiding rescaling the livecell inputs to [-1, 1], while we initialize the model with SAM checkpoints.


anwai98:
This PR is now good to go. Here, we:
1. Refactor the training scripts for all the setups
2. Fix the training and inference (when and when not to rescale the inputs, based on SAM weight initialization)

constantinpape (approved):

PR #181, opened 2023-12-10
Misc updates to transforms, loss and UNETR

constantinpape:


scripts/load_mae_vit.py
```diff
@@ -0,0 +1,13 @@
+from collections import OrderedDict
```
constantinpape, line 1: @anwai98 I have added this script as example for how to load the MAE checkpoint. I will share the link to the checkpoint via RocketChat.

torch_em/model/unetr.py
```diff
@@ -22,37 +23,41 @@ class UNETR(nn.Module):
...
-        use_mae_stats=False,
-        encoder_checkpoint_path=None,
-        final_activation=None,
+        backbone: str = "sam",
+        encoder: str = "vit_b",
+        decoder: Optional[nn.Module] = None,
+        out_channels: int = 1,
+        use_sam_stats: bool = False,
+        use_mae_stats: bool = False,
+        encoder_checkpoint: Optional[Union[str, OrderedDict]] = None,
```
constantinpape, line 59: I have updated the name to reflect that we accept both a path to the checkpoint or the loaded checkpoint (as state dict).

constantinpape (commented):
This contains a couple recent changes, the most important thing is an update to allow passing a state dict to UNETR as well. This makes it easier to change encoder checkpoints, so that they don't have to be resaved and so that we don't need to update the code to parse the checkpoints every time.

@anwai98 with these changes loading the checkpoint from @cdalinghaus still fails with the following error:
```
RuntimeError: Error(s) in loading state_dict for ViT_MAE:                                                                                                                              
        Missing key(s) in state_dict: "head.weight", "head.bias".                                                                                                                      
        size mismatch for pos_embed: copying a param with shape torch.Size([1, 197, 1024]) from checkpoint, the shape in current model is torch.Size([1, 4097, 1024]).
```

scripts/load_mae_vit.py
anwai98: Hi @constantinpape @cdalinghaus. I found the issue, and went ahead and fixed it (https://github.com/constantinpape/torch-em/pull/182).

In short, the ViT is inherited from `timm`, which builds an end-to-end vision transformer for downstream tasks (hence, has the `head` component at the end). The original MAE adaptation just [discards this layer](https://github.com/facebookresearch/mae/blob/main/main_finetune.py#L233-L242), hence I tried here the same and it works. I made the checkpoint more functional to just accept the MAE weights directly (just passing the path to the pretrained MAE weights should do the trick atm). Thanks!

PR #182, opened 2023-12-10
Fix MAE Initialization

anwai98:


torch_em/model/unetr.py
```diff
@@ -41,7 +41,18 @@ def _load_encoder_from_checkpoint(self, backbone, encoder, checkpoint):
...
+                #     - https://github.com/facebookresearch/mae/blob/main/main_finetune.py#L233-L242
+                encoder_state = torch.load(checkpoint)["model"]
+                encoder_state = OrderedDict({
+                    k: v for k, v in encoder_state.items()
+                    if (k != "mask_token" and not k.startswith("decoder"))
+                })
+
+                # let's remove the `head` from our current encoder (as the MAE pretrained don't expect it)
+                current_encoder_state = self.encoder.state_dict()
+                if "head.weight" and "head.bias" in current_encoder_state:
```
constantinpape, line 54: This only checks if `"head.bias" in current_encoder_state`. `"head.weight"` is evaluated as an independent expression and will always evaluate to True.

anwai98: Ah my bad. I just overlooked it. Thanks for pointing. It's fixed now!

constantinpape (approved):

PR #186, opened 2023-12-11
Fix Model Initialization - Update common.py

anwai98:


anwai98:
I fix this in a running workflow already (https://github.com/constantinpape/torch-em/pull/187/commits/8c6648fcaa0dc5cd7254bd1180d45953cb67bbe0), so not required here. I will go ahead and close this.

PR #187, opened 2023-12-12
WIP - Instance Segmentation Experiments

anwai98:


anwai98:
@constantinpape This PR makes the following updates and fixes:
- Updates the postprocessing parameters (subjected to grid search) for distance-based experiments
- Adds `train_by_parts.py`, to check for joint and disjoint training experiments
- Adds a script to check UNETR initialization using different setups
- Makes the `encoder` argument in UNETR modular - to accept pytorch modules or model type as str
- Also makes the use of skip connections flexible.

constantinpape (approved):

PR #192, opened 2023-12-31
Update UNETR to enable resize to longest side

constantinpape:
to enable consistency with SAM and use full encoder input shape rather than naively padding.

This is still a draft to discuss what the best default is and how we can best integrate this with previous code.

### DIFF
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -13,11 +13,6 @@ try:
 except ImportError:
     get_sam_model = None
 
-try:
-    from segment_anything.utils.transforms import ResizeLongestSide
-except ImportError:
-    ResizeLongestSide = None
-
 
 #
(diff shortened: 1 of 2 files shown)


torch_em/model/unetr.py
```diff
@@ -68,10 +73,11 @@ def __init__(
         out_channels: int = 1,
         use_sam_stats: bool = False,
         use_mae_stats: bool = False,
+        resize_input: bool = False,  # TODO should this be the default???
```
anwai98, line 76: I think we can make this the default. This could be our logic to upsample the images (ofc from SAM) that we resize along the longest size first, next pad, and then have the images in the expected dimension. (we also need to keep in mind our plans for dynamic input shapes, but we can discuss that later)

constantinpape:
I went ahead and set resize as the default and also updated the functionality so that we don't need segment-anything for it. I will merge once the tests pass.

anwai98:
Awesome. Thanks Constantin!

PR #196, opened 2024-01-09
Fix AxonDeepSeg

anwai98:


torch_em/data/datasets/axondeepseg.py
```diff
@@ -144,7 +144,7 @@ def get_axondeepseg_dataset(
         if data_fraction is not None:
             assert split is not None
             n_samples = int(len(paths) * data_fraction)
-            paths = paths[:n_samples] if split == "train" else paths[:-n_samples]
+            paths = paths[:n_samples] if split == "train" else paths[-n_samples:]
```
constantinpape, line 147: Good catch, but I think your proposed fix is not correct.
It should be `paths = paths[:n_samples] if split == "train" else paths[n_samples:]`

anwai98: The `n_samples` depends on `data_fraction`. Hence, the combination goes like this:

For training: `..., split="train", data_fraction=0.9`
Hence, n_samples becomes 10 (total volumes) * 0.9 = 9. We then take first 9 samples using `paths[:n_samples]`, which was correct previously.

For validation: `..., split="val", data_fraction=0.1`
Hence, n_samples becomes 10 (total volumes) * 0.1 = 1. We then take the last 1 sample(s) using `paths[-n_samples:]`, which is the current suggestion.

constantinpape: But that introduces two conditions: to get the correct loaders you need to set
```
train_dataset = get_axondeepseg_data(..., split="train", data_fraction=0.9)
val_dataset = get_axondeepseg_data(..., split="val", data_fraction=0.1)
```
That's not great and also not how it's usually handled in the torch_em datasets. It's maybe confusing that this is called `data_fraction` and we should rather call it `val_fraction` (which is also how we handle it elsewhere).

anwai98: Good point. I will take care of the fix then (so that `split` is the only argument this choice is dependent on).

constantinpape (approved):

PR #207, opened 2024-01-30
Update ISBI dataset

anwai98:


anwai98:
GTG from my side! (made the path to install the ISBI volume a little modular)

PR #210, opened 2024-02-03
Update ISBI dataset download

anwai98:


anwai98:
@constantinpape GTG from my side!

constantinpape (approved):

PR #211, opened 2024-02-03
Update UNETR upsampling

anwai98:
I checked the scripts in detail, turns out, the `self.decoder` (the core decoder block) comes from UNet, which already makes use of (bilinear interpolation to scale + conv layer) to perform in-decoder upsampling. It's the features coming from encoder layer(s) which were still making use of convtranspose. Let's see how it works now.


anwai98:
We now have a modular setup, to exchange the upsampling logic (between bilinear interpolation or conv transpose)

anwai98:
@constantinpape GTG from my side! (as mentioned in https://github.com/computational-cell-analytics/micro-sam/pull/379, the default UNETR now makes use of conv transpose)

PR #215, opened 2024-02-09
Refactor CTC dataset

anwai98:


anwai98:
@constantinpape Minor refactoring to:
a) Consider custom split ids
b) Making the download a little modular (to create the path dir, if it doesn't exist)

constantinpape (approved):

PR #216, opened 2024-02-10
Add test set for CTC dataset

anwai98:


anwai98:
After checking the samples, I realised that the test splits don't have labels. I would still add the downloads for them, in case we want to use them for something in future.

torch_em/data/datasets/ctc.py
```diff
@@ -88,7 +100,8 @@ def get_ctc_segmentation_dataset(
     path,
     dataset_name,
     patch_shape,
-    split=None,
+    split,
```
constantinpape, line 103: Let's set the default value for `split` to `"train"`, since the test split does currently not work.

torch_em/data/datasets/ctc.py
```diff
@@ -120,7 +135,8 @@ def get_ctc_segmentation_loader(
     dataset_name,
     patch_shape,
     batch_size,
-    split=None,
+    split,
```
constantinpape, line 138: Set default to `"train"`, see above.

constantinpape (commented):
Let's set the default for split to "train".

torch_em/data/datasets/ctc.py
anwai98: Done!

PR #222, opened 2024-02-23
Update UNETR experiments - setting conv transpose as default

anwai98:


anwai98:
I've update this in another working branch. I'll go ahead and close this for now.

PR #226, opened 2024-03-12
Add ASEM dataset

anwai98:


constantinpape:
The tests are failing because of new bioimageio releases, and it should be enough to wait for everything to catch up.

But I don't want to merge this yet, as the downloads are still excessive and can be done much more efficiently with zarr. I can take care of that.

anwai98:
> The tests are failing because of new bioimageio releases, and it should be enough to wait for everything to catch up.
> 
> But I don't want to merge this yet, as the downloads are still excessive and can be done much more efficiently with zarr. I can take care of that.

Okay, then I continue my experiments on this dataset in this branch (with the current download setup).

constantinpape:
I will go ahead and merge this now. There is still some possibilities to make downloading the data much more efficient, but let's leave this to follow up work. I will create an issue on that.

PR #239, opened 2024-04-12
Update links to the vimunet benchmarking

anwai98:


anwai98:
This is still missing the link to our arXiv preprint.

PR #244, opened 2024-04-26
Add random colors generator for instance plotting

anwai98:


ajinkya-kulkarni:
This commit just popped up on my feed, thanks for it!
BTW, I have tried this before (for diff colored labels for each instance segmented blob), but I realized it often gives very "dark" colored labels.
The code I now use is this, which actually is a part of the StarDist repo.

```
import matplotlib
import colorsys

def random_label_cmap(n=2**16, h=(0, 1), l=(0.4, 1), s=(0.2, 0.8)):
    """
    Generates a random colormap for labeling purposes.

    Parameters:
    n (int): Number of colors in the colormap. Default is 2**16.
    h (tuple): Range of hue values (0 to 1). Default is (0, 1).
    l (tuple): Range of lightness values (0 to 1). Default is (0.4, 1).
    s (tuple): Range of saturation values (0 to 1). Default is (0.2, 0.8).

    Returns:
    matplotlib.colors.ListedColormap: A colormap object with randomly generated colors.

    The function generates colors in HLS (Hue, Lightness, Saturation) space and converts them to RGB.
    The first color in the colormap is always set to black for background or default labeling.
    """

    # Generate random values for hue, lightness, and saturation within specified ranges
    h, l, s = np.random.uniform(*h, n), np.random.uniform(*l, n), np.random.uniform(*s, n)

    # Convert HLS values to RGB values
    cols = np.stack([colorsys.hls_to_rgb(_h, _l, _s) for _h, _l, _s in zip(h, l, s)], axis=0)

    # Set the first color in the colormap to black (useful for background or default labels)
    cols[0] = 0

    # Create a ListedColormap object from the generated RGB values
    random_label_cmap = matplotlib.colors.ListedColormap(cols)

    return random_label_cmap
```
and then just use it as:
```
lbl_cmap = random_label_cmap()
plt.imshow(mask, cmap=lbl_cmap)
```

PR #248, opened 2024-05-09
Add JSRT dataset

anwai98:
This is a dataset for lung segmentation in x-ray images (we used this in our PDA work).
GTG from my side. Thanks!


anwai98:
This PR is GTG from my side!

constantinpape (approved):

PR #249, opened 2024-05-10
Add Montgomery dataset

anwai98:


anwai98:
This is the dataset used in our PDA work as well (for lung segmentation).
There are some minor details to discuss on this, we can discuss this next week!

constantinpape (approved):

PR #256, opened 2024-05-14
Add SIIM ACR Pneumothorax dataset

anwai98:
This is a dataset for segmenting pneumothorax in chest x-rays.

### DIFF
--- a/torch_em/data/image_collection_dataset.py
+++ b/torch_em/data/image_collection_dataset.py
@@ -68,7 +68,8 @@ class ImageCollectionDataset(torch.utils.data.Dataset):
         self.label_images = label_image_paths
         self._ndim = 2
 
-        assert len(patch_shape) == self._ndim
+        if patch_shape is not None:
+            assert len(patch_shape) == self._ndim
         self.patch_shape = patch_shape
 
         self.raw_transform = raw_transform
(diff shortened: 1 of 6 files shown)


anwai98:
In addition, this PR also adds the feature to download files from kaggle using the Kaggle API.

anwai98:
Hi @constantinpape 

This is a nice spot to review a few new features:
- Downloads are now possible from Kaggle (some simple steps in the beginning to make the API work)
- In `ImageCollectionDataset`, it allows now to accept `patch_shape` as None, and load the entire image as it is.
- Added a new transformation function called `ResizeInputs`, which is the `skimage.transform.resize` operation to resize the inputs based on a desired patch shape.

Let me know if you spot something.

torch_em/data/image_collection_dataset.py
```diff
@@ -95,11 +96,16 @@ def ndim(self):
         return self._ndim
 
     def _sample_bounding_box(self, shape):
+        if self.patch_shape is None:
+            patch_shape_for_bb = shape
+        else:
+            patch_shape_for_bb = self.patch_shape
```
constantinpape, line 102: I would also include the `bb_start` logic in this `if-else` and set it to `[0]* len(shape)` if the patch shape is None. 
While I think the current logic woks, it is not very readable.

constantinpape (commented):
- Some minor things can be improved in the handling if `None` patch shape, but looks good overall.
- This logic should also be implemented for the `SegmentationDataset`. Otherwise there will be a disconnect in the functionality.
- Some changes to the resize transformation.

torch_em/transform/generic.py
```diff
@@ -72,6 +72,28 @@ def __call__(self, *inputs):
         return outputs
 
 
+class ResizeInputs:
+    def __init__(self, target_shape, is_label=False):
+        self.target_shape = target_shape
+        self.is_label = is_label
+
+    def __call__(self, inputs):
+        if self.is_label:
```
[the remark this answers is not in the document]
anwai98: Thanks for spotting. Taken care of in a recent commit.

torch_em/data/image_collection_dataset.py
anwai98: Thanks!

anwai98:
Thanks for the mentions!

> Re: This logic should also be implemented for the `SegmentationDataset`. Otherwise there will be a disconnect in the functionality.

I was thinking of doing it in another relevant PR once this is merged, where segmentation dataset is used (ideally for a 3d case like ACDC). Would that work, or should I take care of this in this PR?

constantinpape:
> Would that work, or should I take care of this in this PR?

Yes, that works!

constantinpape (approved):

PR #257, opened 2024-05-15
Add CAMUS dataset

anwai98:
This is a dataset for cardiac structures in 2d (+t) echocardiography images.

@constantinpape - In addition, this PR:
- Makes the `SegmentationDataset` flexible to accept None values now, taking care of returning the entire volume (this data has three dimension, hence tested it for 2d and 3d, works fine)

Let me know if you spot something!

### DIFF
--- a/torch_em/data/segmentation_dataset.py
+++ b/torch_em/data/segmentation_dataset.py
@@ -17,6 +17,9 @@ class SegmentationDataset(torch.utils.data.Dataset):
 
     @staticmethod
     def compute_len(shape, patch_shape):
+        if patch_shape is None:
+            patch_shape = shape
+
         n_samples = int(np.prod([float(sh / csh) for sh, csh in zip(shape, patch_shape)]))
         return n_samples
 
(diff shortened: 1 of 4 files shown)


torch_em/data/segmentation_dataset.py
```diff
@@ -127,10 +139,11 @@ def _get_sample(self, index):
                 if sample_id > self.max_sampling_attempts:
                     raise RuntimeError(f"Could not sample a valid batch in {self.max_sampling_attempts} attempts")
 
-        # squeeze the singleton spatial axis if we have a spatial shape that is larger by one than self._ndim
-        if len(self.patch_shape) == self._ndim + 1:
-            raw = raw.squeeze(1 if self._with_channels else 0)
-            labels = labels.squeeze(1 if self._with_label_channels else 0)
+        if self.patch_shape is not None:
```
constantinpape, line 142: You can combine this into a single condition: `if self.patch_shape is not None and ...`

torch_em/data/segmentation_dataset.py
```diff
@@ -17,6 +17,9 @@ class SegmentationDataset(torch.utils.data.Dataset):
 
     @staticmethod
     def compute_len(shape, patch_shape):
+        if patch_shape is None:
```
constantinpape, line 20: Can you do 
```python
if patch_shape is None:
   return 1
```
instead? That's much easier to read and amounts to the same.

constantinpape (commented):
Only two minor comments, looks good otherwise.

torch_em/data/segmentation_dataset.py
anwai98: Oh yeah, that's simple and much more readable haha. Thanks for the hint.

anwai98:
Thanks for the mentions. The mentions have been taken care of. It's good to go from my side.

constantinpape (approved):

PR #258, opened 2024-05-15
Add ARCADE dataset

anwai98:
This is a dataset for coronary artery segmentation from the ARCADE challenge.

@constantinpape,
The extraction of annotations using pycocotools works as expected for most of the images, except a very few, where it generates annotations from a very different "image_id". I couldn't find the reason why it happens for only handful of images, and not for all the images. (I left the code block in there for now and left some comments on my observation to debug the issue)

For now, I went ahead with the extraction method provided by the challenge organizers, and it works as expected.

If you have time and you spot the cause of the above mentioned ambiguity, let me know. (ideally, would be nice to stick to using pycocotools as the standard for such conversions) Thanks!

PS. The data download is <500MB.


anwai98:
I can't seem to figure out another issue: the image label pairs seem to have a mismatch (not for all images though). I'll investigate this later at some point. I'll leave this open for now!

Ref: https://github.com/cmctec/ARCADE

PR #259, opened 2024-05-15
Add RAVIR dataset

anwai98:
This is a dataset for retinal artery and veins segmentation in infrared reflectance imaging.


anwai98:
This PR is GTG from my side!

PR #260, opened 2024-05-15
Add DRIVE dataset

anwai98:
This is a dataset for retinal vessel segmentation in fundus images.

### DIFF
--- a/torch_em/transform/generic.py
+++ b/torch_em/transform/generic.py
@@ -73,9 +73,10 @@ class Rescale:
 
 
 class ResizeInputs:
-    def __init__(self, target_shape, is_label=False):
+    def __init__(self, target_shape, is_label=False, is_rgb=False):
         self.target_shape = target_shape
         self.is_label = is_label
+        self.is_rgb = is_rgb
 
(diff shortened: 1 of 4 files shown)


torch_em/transform/generic.py
```diff
@@ -73,19 +73,26 @@ def __call__(self, *inputs):
...
             kwargs = {"order": 0,  "anti_aliasing": False}
         else:  # we use the default settings for float data
             kwargs = {}
 
+        if self.is_rgb:
+            assert inputs.ndim == 3 and inputs.shape[0] == 3
+            patch_shape = (3, *self.target_shape)
+        else:
+            patch_shape = self.target_shape
+
```
anwai98, line 92: @constantinpape I chose this scheme to handle resizing with rgb images. Could you double-check the logic here? (I tested with this dataset, and seems to works fine)

constantinpape: Yes, this looks good to me.

anwai98:
Thanks for the feedback. This one's good to go from my side!

constantinpape (approved):

PR #263, opened 2024-05-16
Add Plethora dataset

anwai98:
This is the dataset for pleural effusion and lung segmentation in CT scans.

In addtiion, this PR addresses:
- Adding the `tcia_utils` API to install tcia manifest easily. (WIP)

Looks like it's working. I will test it on the entire dataset and finalize this tomorrow.

### DIFF
--- a/scripts/datasets/check_plethora.py
+++ b/scripts/datasets/check_plethora.py
@@ -1,0 +1,13 @@
+from torch_em.util.debug import check_loader
+from torch_em.data.datasets.medical import get_plethora_loader
+
+
+ROOT = "/media/anwai/ANWAI/data/plethora"
+
+
+def check_plethora():
+    ...
+
+
+if __name__ == "__main__":
+    check_plethora()
--- a/scripts/datasets/check_tcia.py
+++ b/scripts/datasets/check_tcia.py
@@ -4,6 +4,8 @@ from glob import glob
 from natsort import natsorted
 
 import numpy as np
+import pandas as pd
+import nibabel as nib
 import pydicom as dicom
 
 from tcia_utils import nbia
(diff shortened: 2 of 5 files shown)


anwai98:
Hi @constantinpape,

As discussed, I added a new functionality which looks at the `kwargs` if `raw_transform` or `label_transform` are incoming, else applies resize trafo as expected. Let me know if you spot somethinng. This is good to go from my side. Thanks!

torch_em/transform/raw.py
```diff
@@ -227,3 +227,19 @@ def get_default_mean_teacher_augmentations(
         augmentation1=aug1,
         augmentation2=aug2
     )
+
+
+# The functionality below merges incoming and requested transforms to work together.
+class ConcatTransforms:
```
constantinpape, line 233: This is not needed, we have https://github.com/constantinpape/torch-em/blob/main/torch_em/transform/generic.py#L36 already.

torch_em/data/datasets/util.py
```diff
@@ -225,6 +250,33 @@ def add_instance_label_transform(
...
+    """
+    if resize_inputs:
+        assert isinstance(resize_kwargs, dict)
+        patch_shape = None
+
+        raw_trafo = ResizeInputs(target_shape=resize_kwargs["patch_shape"], is_rgb=resize_kwargs["is_rgb"])
+        label_trafo = ResizeInputs(target_shape=resize_kwargs["patch_shape"], is_label=True)
+
+    if "raw_transform" in kwargs:
+        trafo = ConcatTransforms(transform1=kwargs["raw_transform"], transform2=raw_trafo)
```
constantinpape, line 266: Please use `Compose` instead, see comment below.

torch_em/data/datasets/util.py
```diff
@@ -225,6 +250,33 @@ def add_instance_label_transform(
...
+        kwargs["raw_transform"] = trafo
+    else:
+        kwargs["raw_transform"] = raw_trafo
```
constantinpape, line 269: We should use the default raw transform to ensure data standardization here. See https://github.com/constantinpape/torch-em/blob/main/torch_em/segmentation.py#L258

anwai98: Thanks for the mention. I should double check as almost all the medical datasets are missing this. Will take care of these in the incoming PRs (and wherever needed).

anwai98:
Thanks for the feedback, All the mentions have been taken care of now!

PR #267, opened 2024-05-19
Add Kvasir-SEG dataset

anwai98:
This is a dataset for polyp segmentation in colonoscopy images. This is good to go from my side. Thanks!


torch_em/data/datasets/medical/kvasir.py
```diff
@@ -0,0 +1,110 @@
...
+
+    if resize_inputs:
+        raw_trafo = ResizeInputs(target_shape=patch_shape, is_rgb=True)
+        label_trafo = ResizeInputs(target_shape=patch_shape, is_label=True)
+        patch_shape = None
+    else:
+        patch_shape = patch_shape
+        raw_trafo, label_trafo = None, None
+
+    dataset = ImageCollectionDataset(
```
constantinpape, line 83: It's usually better to use `https://github.com/constantinpape/torch-em/blob/main/torch_em/segmentation.py#L188`. This will take care of the defaults for `raw_transform` and `transform` if these are not passed.

torch_em/data/datasets/medical/kvasir.py
```diff
@@ -0,0 +1,110 @@
...
+    The dataset is located at https://datasets.simula.no/kvasir-seg/
+
+    This dataset is from Jha et al. - https://doi.org/10.1007/978-3-030-37734-2_37
+    Please cite it if you use this dataset for a publication.
+    """
+    image_paths, gt_paths = _get_kvasir_paths(path=path, download=download)
+
+    if resize_inputs:
+        raw_trafo = ResizeInputs(target_shape=patch_shape, is_rgb=True)
+        label_trafo = ResizeInputs(target_shape=patch_shape, is_label=True)
```
constantinpape, line 77: It would be better to support additional trafos and concatentate them if `resize_inputs` is given.

torch_em/data/datasets/medical/kvasir.py
anwai98: Yep, this makes total sense. I've moved the dataset to `default_segmentation_dataset` now. Thanks!

torch_em/data/datasets/medical/kvasir.py
anwai98: I've moved the resize inputs to the latest heuristic.

anwai98:
This PR is GTG from my side!

PR #272, opened 2024-05-20
Add CellPose dataset

anwai98:
Need to double check about the RGB images.


torch_em/data/datasets/light_microscopy/cellpose.py
```diff
@@ -0,0 +1,154 @@
...
+        util.unzip(zip_path=zip_path, dst=per_choice_dir, remove=False)
+
+    return data_dir
+
+
+def get_cellpose_dataset(
+    path: Union[os.PathLike, str],
+    split: Literal["train", "test"],
+    patch_shape: Tuple[int, int],
+    choice: Literal["cyto", "cyto2"] = "cyto",
```
constantinpape, line 78: Wouldn't we rather use `cyto2` as the default, since it's the larger dataset?

torch_em/data/datasets/light_microscopy/cellpose.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    return dataset
+
+
+def get_cellpose_loader(
+    path: Union[os.PathLike, str],
+    split: Literal["train", "test"],
+    patch_shape: Tuple[int, int],
+    batch_size: int,
+    choice: Literal["cyto", "cyto2"] = "cyto",
```
constantinpape, line 126: See above.

constantinpape (commented):
Looks good, only wondering about the default dataset.

torch_em/data/datasets/light_microscopy/cellpose.py
anwai98: Looks like `cyto` has more number of images (540) compared to `cyto2` (256).

anwai98: Ah, and `cyto` has provided train-test splits (unlike for `cyto2`, where we only have the train split provided)

anwai98: I like the idea of supporting the possibility of combining both datasets (and is also possible as the image ids do not have any overlaps).

I've extended this feature in the new commit. Thanks for the hint!

constantinpape (approved):

PR #276, opened 2024-05-24
Add Seg-A dataset

anwai98:
This is the dataset for segmentation of aorta in CT scans.


torch_em/util/image.py
```diff
@@ -35,6 +37,9 @@ def load_image(image_path, memmap=True):
         return tifffile.memmap(image_path, mode="r")
     elif tifffile is not None and os.path.splitext(image_path)[1].lower() in (".tiff", ".tif"):
         return tifffile.imread(image_path)
+    elif os.path.splitext(image_path)[1].lower() in (".nrrd"):
```
constantinpape, line 40: I don't think this works / does what is expected, please double check: `(".nrrd")` is the same as `".nrrd"` and then the check is against the elements in the string.
You probably want `(".nrrd",)` instead.

constantinpape (changes requested):

constantinpape (approved):

PR #278, opened 2024-05-26
Add CHAOS dataset

anwai98:
This is the dataset from the CHAOS challenge - for segmentation of abdominal organs in CT and MRI scans.


anwai98:
This dataset is ready and GTG from my side!

constantinpape (approved):

PR #282, opened 2024-05-28
Add SA-Med2D-20M dataset

anwai98:
This is a segmentation dataset for 2d images from various modalities for different tasks.

### DIFF
--- a/torch_em/transform/generic.py
+++ b/torch_em/transform/generic.py
@@ -40,7 +40,7 @@ class Compose:
     def __call__(self, *inputs):
         outputs = self.transforms[0](*inputs)
         for trafo in self.transforms[1:]:
-            outputs = trafo(*outputs)
+            outputs = trafo(outputs)
         return outputs
 
 
--- a/torch_em/transform/generic.py
+++ b/torch_em/transform/generic.py
@@ -34,13 +34,18 @@ class Tile(torch.nn.Module):
 
 # a simple way to compose transforms
 class Compose:
-    def __init__(self, *transforms):
+    def __init__(self, *transforms, is_multi_tensor=True):
         self.transforms = transforms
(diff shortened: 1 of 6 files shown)


anwai98:
@constantinpape Would be good to give this a round of review. Looks good to me, this needs a bit minor checks and refactors from my side, but the core construct is in place. Let me know if you spot something.

PS. I have a few detailed questions, I'll leave comments in those places.

torch_em/transform/generic.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, *transforms):
     def __call__(self, *inputs):
         outputs = self.transforms[0](*inputs)
         for trafo in self.transforms[1:]:
-            outputs = trafo(*outputs)
```
anwai98, line 43: This was the previous state where the outputs from the first trafo were passed as inputs to the next round of trafos using `*outputs`, however when I used this in my workflow, I noticed that `*outputs` splits the outputs (say an array of shape (3, 512, 512)) to a tuple of three different arrays (of shape (512, 512) each). I removed the `*` and it stopped this behaviour. Could you confirm if this is the way-to-go or am I missing something in the transform functions?

For reference: this happens in "raw_transform", where the first transform is the resize inputs transform, and the next is for the default raw transforms.

anwai98: `is_multi_tensor` as another flag to take care of this.

constantinpape (approved):
This looks good to me now. Anything else to add from your side?

anwai98:
I am waiting for the post-processing to complete and to verify the logic of storing the json file. Once that is done, I will ping here!

anwai98:
Hi @constantinpape,

This is ready and good to go from my side. Thanks!

constantinpape (approved):

PR #285, opened 2024-06-03
Add ToothFairy2 dataset

anwai98:
This dataset is for segmentation of 42 different structures in our mouth region imaged using CBCT. This is part of a challenge in MICCAI 2024, would be in our OOD dataset.


anwai98:
For the new version of dataset (ToothFairy2), I decided to merge it with our existing dataloader for the older version (ToothFairy), allowing users to choose between the two versions. This PR is good to go from my side.

constantinpape (approved):

PR #287, opened 2024-06-03
Add PENGWIN dataset

anwai98:
This is the dataset for segmentation of pelvic bone fracture and fragments in CT and X-Ray images. This is from a MICCAI 2024 challenge, and another potential OOD dataset.


anwai98:
This PR finalizes the PENGWIN dataset. GTG from my side!

constantinpape (approved):

PR #288, opened 2024-06-03
Add Acouslic-AI dataset

anwai98:
This is the dataset for segmentation of fetal abdominal circumference in US scans. This could potentially be another OOD dataset.


anwai98:
This dataset is GTG from my side!

constantinpape (approved):

PR #289, opened 2024-06-03
Add MBH Seg dataset

anwai98:
This is a dataset for segmentation of intracranial hemorrhages in non-contrast CT scans. This could potentially be another OOD dataset.


anwai98:
This dataset is GTG from my side! Thanks!

PR #290, opened 2024-06-03
Minor update SA-Med2d 20M dataset

anwai98:
Added an additional logic to choose a fraction of samples per dataset for segmentation, expect the datasets with a small size already.

### DIFF
--- a/torch_em/segmentation.py
+++ b/torch_em/segmentation.py
@@ -206,6 +206,7 @@ def default_segmentation_loader(
     is_seg_dataset=None,
     with_channels=False,
     with_label_channels=False,
+    verify_paths=True,
     **loader_kwargs,
 ):
     ds = default_segmentation_dataset(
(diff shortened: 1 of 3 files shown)


torch_em/segmentation.py
```diff
@@ -250,8 +250,11 @@ def default_segmentation_dataset(
     is_seg_dataset=None,
     with_channels=False,
     with_label_channels=False,
+    verify_paths=True,
 ):
-    check_paths(raw_paths, label_paths)
+    if verify_paths:
+        check_paths(raw_paths, label_paths)
```
anwai98, line 256: @constantinpape Apparently this was the most expensive part (the one in `ImageCollectionDatasets` costs almost nothing). With this, the entire dataset can be loaded in ~30 seconds! ;)

Thanks for the suggestion!

constantinpape: We should probably expose this in `default_segmentation_loader` / `default_segmentation_dataset` too.

constantinpape (commented):
I actually spotted one more change; we should expose the `verify_path` option to the upstream convenience functions.

anwai98:
Thanks for spotting! The argument is now exposed to both `default_segmentation_dataset` and `default_segmentation_loader` now. GTG from my side!

anwai98:
> Should we set `verify_paths=False` when `default_segmentation_dataset` is called in `get_sa_med_dataset`?

Yes, this definitely makes sense to add it to the dataset itself. Thanks for the mention!

PR #294, opened 2024-06-06
Minor updates to multi-gpu training

anwai98:
Hi @constantinpape,

I made some minor changes to the DDP-based training to fit our SAM finetuning. ~There are a few issues in handling `mixed_precision`, would be good to take a look at it.~


anwai98:
For reference, this is the error I get:

`AssertionError: No inf checks were recorded for this optimizer.`

anwai98:
@constantinpape This is GTG from my side.

torch_em/multi_gpu_training.py
```diff
@@ -33,37 +33,82 @@ def _create_data_loader(ds_callable, ds_kwargs, loader_kwargs, world_size, rank)
     return loader
 
 
+class DDP(DistributedDataParallel):
```
constantinpape, line 36: Can you explain why this is necessary?

torch_em/multi_gpu_training.py
```diff
@@ -33,37 +33,82 @@ def _create_data_loader(ds_callable, ds_kwargs, loader_kwargs, world_size, rank)
...
-    loader_kwargs, iterations, **kwargs
+    model_callable,
+    model_kwargs,
+    train_dataset_callable,
+    train_dataset_kwargs,
+    val_dataset_callable,
+    val_dataset_kwargs,
+    loader_kwargs,
+    iterations,
+    find_unused_parameters=False,
```
constantinpape, line 105: From what I understand this is important to make the training work for more complex models. So should we set the default to `True`?

constantinpape (commented):
Looks good overall, only two minor comments.

torch_em/multi_gpu_training.py
anwai98: Ah yes, to access the original model's functionalities (say `self.model.image_embeddings_oft` for instance), the calls from `DDP` wrapped models are `self.model.module.image_embeddings_oft`. I decided to generalize this via another class to automate this, rather changing it on the model-level calls.

torch_em/multi_gpu_training.py
anwai98: Yep, you are correct. I've set the default value to True now. Thanks for the suggestion!

torch_em/multi_gpu_training.py
constantinpape: I see! That makes sense. This is a good design! Can you add a short comment on this?

anwai98: Thanks! Added a docstring to the wrapper now!

constantinpape (approved):

PR #295, opened 2024-06-10
Minor fix to platynereis cilia dataset

anwai98:
This PR fixes one minor issue: passing the `rois` selected by the user for the volumes. (also adds a script for testing dataloaders for the next round of LoRA experiments, not so relevant)


anwai98:
@constantinpape This is GTG from my side! Thanks!

Issue #320, opened 2024-07-10
add new label transform


lufre1:
add a new label transform function or class that can handle min_size of instances

PR #322, opened 2024-07-11
added label transform class encorporating min_size of labels

lufre1:
[issue](https://github.com/constantinpape/torch-em/issues/320)

### DIFF
--- a/torch_em/transform/label.py
+++ b/torch_em/transform/label.py
@@ -37,6 +37,25 @@ def label_consecutive(labels, with_background=True):
     return seg
 
 
+class MinSizeLabelTransform:
+    def __init__(self, min_size=None, ndim=None, ensure_zero=False):
+        self.min_size = min_size
+        self.ndim = ndim
+        self.ensure_zero = ensure_zero
+
+    def __call__(self, labels):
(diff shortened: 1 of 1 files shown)


torch_em/transform/label.py
```diff
@@ -37,6 +37,25 @@ def label_consecutive(labels, with_background=True):
...
+class MinSizeLabelTransform:
+    def __init__(self, min_size=None, ndim=None, ensure_zero=False):
+        self.min_size = min_size
+        self.ndim = ndim
+        self.ensure_zero = ensure_zero
+
+    def __call__(self, labels):
+        components = connected_components(labels, ndim=self.ndim, ensure_zero=self.ensure_zero)
+        if self.min_size is not None:
+            for component in np.unique(components)[1:]:  # Skip background (label 0)
```
constantinpape, line 49: This is inefficient and your labels are not consecutively labeled anymore after filtering. You can do the following instead:
```python
ids, sizes = np.unique(components, return_counts=True)
filter_ids = ids[sizes < self.min_size]
components[np.isin(components, filter_ids)] = 0
components, _, _ = skimage.segmentation.relabel_sequential(components)
```

lufre1: updated the code

constantinpape (approved):

PR #323, opened 2024-07-11
Mininstancesampler update

lufre1:
@constantinpape

### DIFF
--- a/torch_em/data/sampler.py
+++ b/torch_em/data/sampler.py
@@ -1,6 +1,5 @@
 import numpy as np
 from typing import List
-import skimage.segmentation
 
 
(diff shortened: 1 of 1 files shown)


torch_em/data/sampler.py
```diff
@@ -85,13 +85,18 @@ class MinInstanceSampler:
     def __init__(
         self,
         min_num_instances: int = 2,
-        p_reject: float = 1.0
+        p_reject: float = 1.0,
+        min_size: int = None
```
constantinpape, line 89: This should be `min_size: Optional[int] = None`. And then you need to include the `Optional` import on top.

PR #326, opened 2024-07-17
Add OmniPose dataset

anwai98:


scripts/datasets/light_microscopy/check_omnipose.py
```diff
@@ -0,0 +1,21 @@
+from torch_em.util.debug import check_loader
+from torch_em.data import MinInstanceSampler
+from torch_em.data.datasets.light_microscopy import get_omnipose_loader
+
+
+ROOT = "/media/anwai/ANWAI/data/omnipose"
```
constantinpape, line 6: Adapt to the new ROOT logic. See also comments in https://github.com/constantinpape/torch-em/pull/330

torch_em/data/datasets/light_microscopy/omnipose.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+    return dataset
+
+
+def get_omnipose_loader(
+    path: Union[os.PathLike, str],
+    patch_shape: Tuple[int, int],
+    batch_size: int,
+    split: Literal["train", "test"],
+    data_choice: Literal["bact_fluor", "bact_phase", "worm", "worm_high_res"],
```
constantinpape, line 122: Would it make sense to add `None` as (the default?) option and then load all the different data types?
(I am not sure, we don't have to.)

PR #329, opened 2024-07-29
Add ResizeLongestSideInputs transform

anwai98:


torch_em/transform/generic.py
```diff
@@ -105,6 +106,52 @@ def __call__(self, inputs):
         return inputs
 
 
+class ResizeLongestSideInputs:
+    def __init__(self, target_shape, target_length=512, is_label=False, is_rgb=False):
```
constantinpape, line 110: Why do we need the `target_length` here?

torch_em/transform/generic.py
```diff
@@ -105,6 +106,52 @@ def __call__(self, inputs):
...
+        self.target_shape = target_shape
+        self.target_length = target_length
+        self.is_label = is_label
+        self.is_rgb = is_rgb
+
+    def _get_preprocess_shape(self, oldh, oldw):
+        """Inspired from Segment Anything."""
+        scale = self.target_length * 1.0 / max(oldh, oldw)
```
constantinpape, line 118: Can't this just be derived from `target_shape`?

constantinpape (commented):
I think the logic here is somehow wrong. Having `target_length` independent of `target_shape` does not make sense to me.

And in general, I think this is functionality for which tests should be added.

torch_em/transform/generic.py
anwai98: Yep, you are correct. It's not necessary to have this at all. I'll take care of this.

torch_em/transform/generic.py
anwai98: True. On it!

torch_em/transform/generic.py
```diff
@@ -105,6 +106,65 @@ def __call__(self, inputs):
...
+    def __init__(self, target_shape, is_label=False, is_rgb=False):
+        self.target_shape = target_shape
+        self.is_label = is_label
+        self.is_rgb = is_rgb
+
+    def _get_target_length(self):
+        h, w = self.target_shape[-2], self.target_shape[-1]
+
+        # We currently support resize feature for square-shaped target shape only.
+        if not h == w:
```
constantinpape, line 119: `if h != w` is more readable.

constantinpape: And I would do this check already in the `__init__`.

test/transform/test_generic.py
```diff
@@ -34,6 +34,18 @@ def _test_tile_impl(ndim, reps):
         actual = tile_aug(a)
         assert actual.shape == expected.shape
 
+    def test_resize_longest_inputs(self):
+        input_shape = (520, 704)
+        target_shape = (512, 512)
```
constantinpape, line 39: I would rather test this for a few different input and target shapes.
E.g.
```python
input_shapes = [(520, 704), (256, 384), ...]
target_shapes =  [(512, 512), (1024, 1024), ...]
```
And then have a loop that runs the test over those. And then check different cases, like target shape smaller and bigger than input shape.

constantinpape (commented):
Looks good now, only a few minor things.

test/transform/test_generic.py
anwai98: Added more combinations for it!

constantinpape (approved):

PR #330, opened 2024-07-29
Add padding to SegmentationDataset

anwai98:


scripts/datasets/light_microscopy/check_mouse_embryo.py
```diff
@@ -1,20 +1,31 @@
-import os
-import sys
-
-from torch_em.data.datasets import get_mouse_embryo_loader
 from torch_em.util.debug import check_loader
+from torch_em.data.datasets import get_mouse_embryo_loader
```
constantinpape, line 2: Can you please reset this file? There is a new logic to the scripts to set the ROOT directory globally in `../util.py`. Please use this logic to set your appropriate root dir rather than changing it in the scripts, as this will otherwise lead to many meaningless changes.

scripts/datasets/light_microscopy/check_mouse_embryo.py
```diff
@@ -1,20 +1,31 @@
...
-sys.path.append("..")
 
+ROOT = "/media/anwai/ANWAI/data/mouse_embryo"
 
-def check_mouse_embryo():
-    from util import ROOT
 
-    data_root = os.path.join(ROOT, "mouse_embryo")
-    loader = get_mouse_embryo_loader(data_root, "nuclei", "train", (8, 512, 512), 1, download=True)
+def check_mouse_embryo():
```
constantinpape, line 8: The changes here are ok, but please change the ROOT logic. See above.

torch_em/data/segmentation_dataset.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         ndim: Optional[int] = None,
         with_channels: bool = False,
         with_label_channels: bool = False,
+        with_padding_per_patch: bool = False,
```
constantinpape, line 45: I would just call this `with_padding`.

constantinpape: And we can set the default to `True`.

torch_em/data/image_collection_dataset.py
constantinpape, line 1: We should also add the `with_padding` argument here, so that this behaves the same as `SegmentationDataset`.

torch_em/segmentation.py
```diff
@@ -253,6 +253,7 @@ def default_segmentation_dataset(
     with_channels=False,
     with_label_channels=False,
     verify_paths=True,
+    with_padding_per_patch=False,
```
constantinpape, line 256: Update the name (see above), set default to True and pass it to ImageCollectionDataset as well (after adding it there as well).

torch_em/data/image_collection_dataset.py
anwai98: Ah I see, I'll make it an optional component then (at the moment, it takes place all the time)

scripts/datasets/light_microscopy/check_mouse_embryo.py
anwai98: I see, didn't spot it carefully. Thanks for mentioning. I'll make the other datasets from my side consistent as well.

torch_em/data/image_collection_dataset.py
anwai98: Done. Thanks for mentioning!

torch_em/data/segmentation_dataset.py
```diff
@@ -123,11 +125,22 @@ def _sample_bounding_box(self):
...
             raise RuntimeError("SegmentationDataset has not been properly deserialized.")
+
         bb = self._sample_bounding_box()
         bb_raw = (slice(None),) + bb if self._with_channels else bb
         bb_labels = (slice(None),) + bb if self._with_label_channels else bb
         raw, labels = self.raw[bb_raw], self.labels[bb_labels]
 
+        # Padding the patch to match the expected input shape.
+        if self.patch_shape is not None and self.with_padding:
+            if any(sh < psh for sh, psh in zip(raw.shape, self.patch_shape)):
```
constantinpape, line 136: This logic will fail if we have channels. This is also why we use `_ensure_patch_shape` in `ImageCollectionDataset`. I would suggest that we refactor `_ensure_patch_shape` into `torch_em.util` so that we can use this in both datasets.

constantinpape (commented):
I noticed another issue. Otherwise the refactoring looks good.

torch_em/data/segmentation_dataset.py
anwai98: This is updated now!

constantinpape (approved):

PR #332, opened 2024-08-02
Add resize transform support for 3d

anwai98:


torch_em/transform/generic.py
```diff
@@ -138,12 +138,15 @@ def __call__(self, inputs):
         # Let's get the new shape with the longest side equal to the target length.
         new_shape = self._get_preprocess_shape(inputs.shape[-2], inputs.shape[-1])
 
-        if self.is_rgb:
+        if self.is_rgb:  # for rgb inputs, we assume channels first
             assert inputs.ndim == 3 and inputs.shape[0] == 3
             patch_shape = (3, *new_shape)
+        elif inputs.ndim == 3:  # for 3d inputs, we assume channels first
```
constantinpape, line 144:
```suggestion
        elif inputs.ndim == 3:  # for 3d inputs, we do not resize along the first (=z) axis

File: torch_em/transform/generic.py [hunk 2]
```diff
@@ -153,7 +156,7 @@ def __call__(self, inputs):
         pad_width = (
             (ceil(pad_width[0]), floor(pad_width[0])), (ceil(pad_width[1]), floor(pad_width[1]))
         )
-        if self.is_rgb:
+        if self.is_rgb or inputs.ndim == 3:  # in either cases, we assume channels first.
```
constantinpape, line 159:
```suggestion
        if self.is_rgb or inputs.ndim == 3:  # we do not pad across the first axis (= channel or z-axis) for rgb or 3d inputs
```

constantinpape (approved):
Thanks @anwai98 ! This is implementing exactly what we discussed :). (Though I may have overlooked some corner case, but I think you did test this already :) ). I just updated two comments, which were a bit misleading (in the case of 3d data we do not assume to have a channel axis, but instead have the spatial z-axis)

anwai98:
Thanks @constantinpape. I did test it for all existing use-cases with SegmentationDataset, however the z-dim padding still is something we could handle in future a bit better.

PR #334, opened 2024-08-06
changed order of segmentationdataset's get_sample function logic: sam…

lufre1:
…pler first, then ensure patch_shape


lufre1:
@constantinpape

constantinpape (approved):
This makes sense. Thanks!!

Issue #341, opened 2024-09-20
Issues with PlantSeg dataset


anwai98:
I have observed the following issues (under the `Root` subset):
- Some test volumes have a few empty patches towards the end of the stack
- Some volumes seem to be repeated across splits (eg. `Movie1_t00045_crop_gt.h5`). It could surely be a splitting / naming issue, need to double check this.

PS. Also double-check the aforementioned issues in other data subset in PlantSeg.

anwai98:
Okay, I verified the above points, and:
- Re: empty patches or artifacts in some volumes - seems like it's the case for quite a few volumes in root, and some in ovules as well.
- Re: repeated volumes in train and test for root - it seems like both the volumes are same. I went ahead and removed it as well.

I went ahead and fixed both issues in the aforementioned PR.

PR #344, opened 2024-09-22
Fix plantseg volumes - remove slices with artifacts

anwai98:
Fixes https://github.com/constantinpape/torch-em/issues/341

### DIFF
--- a/torch_em/data/datasets/light_microscopy/plantseg.py
+++ b/torch_em/data/datasets/light_microscopy/plantseg.py
@@ -91,19 +91,25 @@ def _fix_inconsistent_volumes(data_path, name, split):
     for vol_path in tqdm(file_paths, desc="Fixing inconsistencies in volumes"):
         fname = os.path.basename(vol_path)
 
-        if fname not in CROPPING_VOLUMES:
-            continue
-
         if fname == "Movie1_t00045_crop_gt.h5":  # to avoid duplicated volumes in 'train' and 'test'.
             os.remove(vol_path)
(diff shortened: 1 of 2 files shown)


torch_em/data/datasets/light_microscopy/plantseg.py
```diff
@@ -56,6 +83,36 @@
...
+def _fix_inconsistent_volumes(data_path, name, split):
+    file_paths = glob(os.path.join(data_path, "*.h5"))
+    if name not in ["root", "ovules"] and split not in ["train", "val"]:
+        return
+
+    for vol_path in tqdm(file_paths, desc="Fixing inconsistencies in volumes"):
+        fname = os.path.basename(vol_path)
+
+        # avoid duplicated volumes in 'train' and 'test'.
+        if fname == "Movie1_t00045_crop_gt.h5" and (name == "plantseg" and split == "train"):
```
constantinpape, line 95: I don't see how `name == "plantseg"` can occur here. The only valid names are `"root", "ovules", "nuclei"`. Do you mean `"root"`?

constantinpape (commented):
There is an inconsistency with one of the names. Looks good otherwise.

torch_em/data/datasets/light_microscopy/plantseg.py
anwai98: Oops, yes. That's a mistake from my side. Thanks for spotting!

constantinpape (approved):

PR #348, opened 2024-09-30
Extend functionality to obtain original inputs from resized inputs

anwai98:
Hi @constantinpape,

I decided to extend our transform, to support re-transforming the reshaped inputs to the original size. I decided upon this design in favor of avoiding to store different shapes to unpad and resize the transformed inputs. Let me know if you think otherwise. Thanks!

### DIFF
--- a/torch_em/transform/generic.py
+++ b/torch_em/transform/generic.py
@@ -136,7 +136,7 @@ class ResizeLongestSideInputs:
 
     def convert_transformed_inputs_to_original_shape(self, resized_inputs):
         if not hasattr(self, "pre_pad_shape"):
-            raise AttributeError(
+            raise RuntimeError(
                 "'convert_transformed_inputs_to_original_shape' is only valid after the '__call__' method has run."
             )
 
--- a/torch_em/transform/generic.py
+++ b/torch_em/transform/generic.py
@@ -134,7 +134,12 @@ class ResizeLongestSideInputs:
         newh = int(newh + 0.5)
         return (newh, neww)
 
-    def _convert_transformed_inputs_to_original_shape(self, resized_inputs):
+    def convert_transformed_inputs_to_original_shape(self, resized_inputs):
+        if not hasattr(self, "pre_pad_shape"):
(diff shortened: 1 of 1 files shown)


torch_em/transform/generic.py
```diff
@@ -129,11 +134,18 @@ def _get_preprocess_shape(self, oldh, oldw):
         newh = int(newh + 0.5)
         return (newh, neww)
 
+    def _convert_transformed_inputs_to_original_shape(self, resized_inputs):
```
constantinpape, line 137: I would remove the `_` from the name, since this is an externally facing function.

torch_em/transform/generic.py
```diff
@@ -129,11 +134,18 @@ def _get_preprocess_shape(self, oldh, oldw):
...
+        # First step is to remove the padded region
+        inputs = resized_inputs[tuple(self.pre_pad_shape)]
```
constantinpape, line 139: I would add a check here that `pre_pad_shape` exists. and raise an error if it doesn't exists that informs that this is only valid to be used after `__call__` has run.
You can check this via
```python
if not hasattr(self, "pre_pad_shape"):
   ...
```

constantinpape (commented):
I think this is a very good idea! Let's expose it properly.

torch_em/transform/generic.py
anwai98: Ahha makes sense. Thanks for the feedback!

torch_em/transform/generic.py
```diff
@@ -129,11 +134,23 @@ def _get_preprocess_shape(self, oldh, oldw):
         newh = int(newh + 0.5)
         return (newh, neww)
 
+    def convert_transformed_inputs_to_original_shape(self, resized_inputs):
+        if not hasattr(self, "pre_pad_shape"):
+            raise AttributeError(
```
constantinpape, line 139: `AttributeError` doesn't really make sense for this. I would rather throw a `RuntimeError`.

Issue #350, opened 2024-10-01
Check HPA dataset processing


anwai98:
I think the design for processing images could change a bit.

Reason: The part for [`_process_image`](https://github.com/constantinpape/torch-em/blob/main/torch_em/data/datasets/light_microscopy/hpa.py#L261-L287) processes downloaded inputs in a selective-channel fashion (say compose the h5 files with 'nuclei' channel only - as the user requested for it). This might create an issue later to look for new channels. The only solution then is to remove the dataset and download it with the desired channel specification.

Solution: Store each channel into separate groups under `raw`: (say `raw/microtubules`, `raw/protein`, `raw/nuclei` and `raw/er`)

Do we think this is a good solution or should we opt for another way?

cc: @constantinpape

constantinpape:
I think this is a good solution, but we then also need to combine the dataset to stack the selected channels on the fly.

anwai98:
Updated at https://github.com/constantinpape/torch-em/pull/351.

PR #351, opened 2024-10-01
Update design for storing input channels in HPA dataset

anwai98:
Takes care of https://github.com/constantinpape/torch-em/issues/350.

I made a minor addition to the way we treat multiple groups (i.e. `raw_key` / `label_key`) passed by the user and then likewise stack inputs from multiple groups. Let me know if you think there is a better way to handle this.

### DIFF
--- a/scripts/datasets/light_microscopy/check_hpa.py
+++ b/scripts/datasets/light_microscopy/check_hpa.py
@@ -10,7 +10,14 @@ sys.path.append("..")
 def check_hpa():
     from util import ROOT
 
-    loader = get_hpa_segmentation_loader(os.path.join(ROOT, "hpa"), "train", (512, 512), 1, download=True)
+    loader = get_hpa_segmentation_loader(
+        path=os.path.join(ROOT, "hpa"),
+        split="train",
+        patch_shape=(1024, 1024),
+        batch_size=1,
+        channels=["protein"],
+        download=True,
+    )
     check_loader(loader, 8, instance_labels=True)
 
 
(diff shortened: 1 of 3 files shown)


torch_em/data/datasets/light_microscopy/hpa.py
```diff
@@ -370,7 +370,7 @@ def get_hpa_segmentation_dataset(
     Returns:
        The segmentation dataset.
     """
-    get_hpa_segmentation_data(path, download, channels, n_workers_preproc)
+    get_hpa_segmentation_data(path, download, n_workers_preproc)
```
constantinpape, line 373: We should add a check here to ensure that `channels` passed by the user are valid.

scripts/datasets/light_microscopy/check_hpa.py
```diff
@@ -10,7 +10,14 @@
 def check_hpa():
     from util import ROOT
 
-    loader = get_hpa_segmentation_loader(os.path.join(ROOT, "hpa"), "train", (512, 512), 1, download=True)
+    loader = get_hpa_segmentation_loader(
+        path=os.path.join(ROOT, "hpa"),
+        split="train",
+        patch_shape=(1024, 1024),
+        batch_size=1,
+        channels=["protein"],
```
constantinpape, line 18: Have you checked that this works when using multiple channels?

constantinpape (commented):
This is a very nice solution! Only two minor comments.

scripts/datasets/light_microscopy/check_hpa.py
anwai98: Yes, it works as expected!

constantinpape (approved):

Issue #356, opened 2024-10-04
Set non blocking to True for device allocation in lazy loading 


anwai98:
I've observed in some of my experiments that using `x.to(non_blocking=True)` for moving inputs to GPU memory is faster. It would make sense to switch the device allocation of inputs to this.

NOTE: For this, we need to provide `pin_memory=True` to the torch dataloaders.

Ref: https://github.com/constantinpape/torch-em/blob/main/torch_em/trainer/default_trainer.py#L670

PR #357, opened 2024-10-04
Set non_blocking to True for tensor device allocation

anwai98:
This PR adds `non_blocking=True` to tensor pins to `cuda` to enable asynchronous copies of objects, leading to a minor speed up.

NOTE: This only works when items originate from pinned memory, i.e. works only when the argument `pin_memory=True` is passed to the dataloader.

Reference: https://pytorch.org/docs/stable/notes/cuda.html#cuda-memory-pinning

This should in theory speed up respective trainings as well. GTG from my side. Thanks!

Closes https://github.com/constantinpape/torch-em/issues/356.

### DIFF
--- a/torch_em/segmentation.py
+++ b/torch_em/segmentation.py
@@ -318,8 +318,6 @@ def default_segmentation_dataset(
 
 def get_data_loader(dataset: torch.utils.data.Dataset, batch_size: int, **loader_kwargs) -> torch.utils.data.DataLoader:
     pin_memory = loader_kwargs.pop("pin_memory", True)
-    print(pin_memory)
-    breakpoint()
     loader = torch.utils.data.DataLoader(dataset, batch_size=batch_size, pin_memory=pin_memory, **loader_kwargs)
     # monkey patch shuffle attribute to the loader
     loader.shuffle = loader_kwargs.get("shuffle", False)
--- a/torch_em/segmentation.py
+++ b/torch_em/segmentation.py
@@ -316,8 +316,11 @@ def default_segmentation_dataset(
     return ds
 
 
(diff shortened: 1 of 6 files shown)


constantinpape (commented):
I think that the changes are good, but maybe we should also set `pin_memory=True` by default in order to actually get a speed-up. This could be easily added in [get_data_loader](https://github.com/constantinpape/torch-em/blob/main/torch_em/segmentation.py#L319):
```python
pin_memory = loader_kwargs.pop("pin_memory", True)
loader = torch.utils.DataLoader(dataset, batch_size=batch_size, pin_memory=pin_memory, **loader_kwargs)
...
```
What do you think @anwai98 ?

anwai98:
> I think that the changes are good, but maybe we should also set `pin_memory=True` by default in order to actually get a speed-up. 

Yes, I do agree that we should switch the defaults to `True`. I've taken care of it in the latest commit. Thanks for the suggestion!

torch_em/segmentation.py
```diff
@@ -316,8 +316,11 @@ def default_segmentation_dataset(
     return ds
 
 
-def get_data_loader(dataset: torch.utils.data.Dataset, batch_size, **loader_kwargs) -> torch.utils.data.DataLoader:
-    loader = torch.utils.data.DataLoader(dataset, batch_size=batch_size, **loader_kwargs)
+def get_data_loader(dataset: torch.utils.data.Dataset, batch_size: int, **loader_kwargs) -> torch.utils.data.DataLoader:
+    pin_memory = loader_kwargs.pop("pin_memory", True)
+    print(pin_memory)
+    breakpoint()
```
constantinpape: It seems like you missed some debugging output ;)

anwai98: Ah yes, sorry! I removed it in the latest commit. Thanks for spotting!

constantinpape (approved):

PR #358, opened 2024-10-05
Update Probabilstic UNet model

anwai98:


torch_em/model/probabilistic_unet.py
```diff
@@ -282,16 +282,18 @@ class ProbabilisticUNet(nn.Module):
 
     The following elements are initialized to get our desired network:
     input_channels: the number of channels in the image (1 for grayscale and 3 for RGB)
-    num_classes: the number of classes to predict
+    output_channels: the number of channels to predict.
+    num_classes: the number of classes (raters) for the posterior
```
anwai98, line 286: num_raters

anwai98:
I'll close this and move it to the primary repo, instead of my fork (would be easier for me to debug with Laura ;)) (have some major findings here why it didn't work before as well, just need to quantify it)

PR #367, opened 2024-10-09
Add function in datasets to get input paths

anwai98:
This PR extends support for the possibility of having paths to input images / volumes. This is in favor of https://github.com/computational-cell-analytics/micro-sam/pull/728 to avoid writing additional scripts to extracts paths split-wise / data partition wise.

It's a WIP, I'll put this in review once the required (ideally all) datasets have the support for the mentioned feature.

### DIFF
--- a/torch_em/data/datasets/light_microscopy/deepbacs.py
+++ b/torch_em/data/datasets/light_microscopy/deepbacs.py
@@ -100,7 +100,7 @@ def get_deepbacs_data(path: Union[os.PathLike, str], bac_type: str, download: bo
     return data_folder
 
 
-def _get_paths(path, bac_type, split):
+def _get_deepbacs_paths(path, bac_type, split):
     # the bacteria types other than mixed are a bit more complicated so we don't have the dataloaders for them yet
     # mixed is the combination of all other types
     if split == "train":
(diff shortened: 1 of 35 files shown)


torch_em/data/datasets/light_microscopy/deepbacs.py
```diff
@@ -100,7 +100,7 @@ def get_deepbacs_data(path: Union[os.PathLike, str], bac_type: str, download: bo
     return data_folder
 
 
-def _get_paths(path, bac_type, split):
+def _get_deepbacs_paths(path, bac_type, split):
```
constantinpape, line 103: If we make this a standard across dataset then we can remove the `_` and add docstrings.

constantinpape (commented):
I think this is a good idea. If we standardize it we can also document it though.

torch_em/data/datasets/light_microscopy/deepbacs.py
anwai98: Thanks for the suggestion. I decided to go ahead and implement this for all microscopy datasets now.

constantinpape (approved):
Looks good high-level. I didn't check all changes in detail. We'll find issues when using the datasets.

PR #369, opened 2024-10-10
Add function to return input paths in histopathology datasets

anwai98:
This PR adds `get_<DATASET>_paths` functionality to histopathology datasets. It's GTG from my side!

### DIFF
--- a/torch_em/data/datasets/light_microscopy/covid_if.py
+++ b/torch_em/data/datasets/light_microscopy/covid_if.py
@@ -93,7 +93,10 @@ def get_covid_if_dataset(
         sample_range: Id range of samples to load from the training dataset.
         target: The segmentation task. Either 'cells' or 'nuclei'.
         download: Whether to download the data if it is not present.
-        
+        offsets: Offset values for affinity computation used as target.
+        boundaries: Whether to compute boundaries as the target.
+        binary: Whether to use a binary segmentation target.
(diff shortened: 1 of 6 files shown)


torch_em/data/datasets/light_microscopy/covid_if.py
```diff
@@ -93,10 +93,7 @@ def get_covid_if_dataset(
         sample_range: Id range of samples to load from the training dataset.
         target: The segmentation task. Either 'cells' or 'nuclei'.
         download: Whether to download the data if it is not present.
-        offsets: Offset values for affinity computation used as target.
-        boundaries: Whether to compute boundaries as the target.
-        binary: Whether to use a binary segmentation target.
-        kwargs: Additional keyword arguments for `torch_em.default_segmentation_dataset`.
```
constantinpape, line 99: This looks wrong.  These parameters are still there, so we shouldn't remove the doc.

constantinpape (commented):
The changes in histopatho look good, but there seems to be an accidental change in covid-if.

torch_em/data/datasets/light_microscopy/covid_if.py
anwai98: Ah yeah my bad. It should be restored now. Thanks for spotting!

constantinpape (approved):

PR #370, opened 2024-10-10
Add EMNeuron dataset

anwai98:


anwai98:
This PR adds a large dataset for neuron segmentation in vEM datasets. Looks quite interesting with an appreciable amount of domain variability across the dataset. GTG from my side, thanks!

PS: Auto download is not supported for the dataset (I've added instructions at `get_emneuron_data` on how to download the data.

Ref: https://papers.miccai.org/miccai-2024/677-Paper0518.html

constantinpape (approved):

Issue #371, opened 2024-10-12
Another large histopatho dataset


constantinpape:
https://www.sciencedirect.com/science/article/pii/S0010482521001438

cc @anwai98

PR #372, opened 2024-10-12
Add CryoNuSeg dataset

anwai98:
This PR adds CryoNuSeg for nuclei segmentation in histology images. The data definitely looks quite interesting and relevant to us. (thanks for spotting)
GTG from my side!

Closes https://github.com/constantinpape/torch-em/issues/371.


constantinpape (changes requested):
It looks like the actual dataset is not added. Probably you just forgot the `git add` for it.

anwai98:
Ah my bad. I totally forgot to add it. Everything should be as expected now. Thanks for spotting!

PR #379, opened 2024-10-14
378 add ensuring patch shape for raw dataset

lufre1:
@constantinpape 
added the ensure patch shape for raw dataset without labels

### DIFF
--- a/torch_em/data/raw_dataset.py
+++ b/torch_em/data/raw_dataset.py
@@ -112,7 +112,7 @@ class RawDataset(torch.utils.data.Dataset):
                 raw=raw,
                 labels=None,
                 patch_shape=self.patch_shape,
-                with_channels=self._with_channels
+                have_raw_channels=self._with_channels
             )
         # squeeze the singleton spatial axis if we have a spatial shape that is larger by one than self._ndim
         if len(self.patch_shape) == self._ndim + 1:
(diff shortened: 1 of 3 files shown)


torch_em/data/raw_dataset.py
```diff
@@ -107,7 +107,12 @@ def _get_sample(self, index):
                 sample_id += 1
                 if sample_id > self.max_sampling_attempts:
                     raise RuntimeError(f"Could not sample a valid batch in {self.max_sampling_attempts} attempts")
-
+        if self.patch_shape is not None:
+            raw = ensure_patch_shape(
+                raw=raw,
+                labels=None,
+                patch_shape=self.patch_shape,
```
constantinpape, line 114: I think we also have to pass
```
have_raw_channels=self._with_channels
```
here. Otherwise it will not work if the data has channels.

constantinpape (commented):
I think a function argument is missing and that this is also why the test  fails: ` ERROR: test_with_channels (data.test_raw_dataset.TestRawDataset.test_with_channels)`

torch_em/data/raw_dataset.py
```diff
@@ -107,7 +107,13 @@ def _get_sample(self, index):
...
+                with_channels=self._with_channels
```
constantinpape, line 115: The argument is called `have_raw_channels`. See also my previous comment.

lufre1: corrected it accordingly

constantinpape (approved):

PR #382, opened 2024-10-15
add scheduled confidence threshold within the ScheduledPseudoLabeler class

Marei33:


torch_em/util/util.py
```diff
@@ -283,6 +283,7 @@ def load_model(checkpoint, model=None, name="best", state_key="model_state", dev
             ckpt = checkpoint
 
         state = torch.load(ckpt, map_location=device)[state_key]
+
```
Marei33, line 286: Check weights_only argument

torch_em/self_training/mean_teacher.py
```diff
@@ -235,6 +235,7 @@ def _train_epoch_unsupervised(self, progress, forward_context, backprop):
                 )
                 lr = [pm["lr"] for pm in self.optimizer.param_groups][0]
                 self.logger.log_lr(self._iteration, lr)
+                self.logger.log_ct(self._iteration, self.pseudo_labeler.confidence_threshold)
```
constantinpape, line 238: Maybe we need to check here that the confidence_threshold is not None before logging?

constantinpape (commented):
Hi @Marei33,
I had a quick look, and the overall design looks very good!

However, there is an issue with a test that seems to be caused by your changes:
```
ERROR: test_mean_teacher_unsupervised (self_training.test_mean_teacher.TestMeanTeacher.test_mean_teacher_unsupervised)
```
Could you check out why this fails? 
You can run the test locally by going to `test/self_training` and then running
```
python test_mean_teacher.py
```

torch_em/self_training/mean_teacher.py
Marei33: Yes, I did exactly that. Now it only logs the threshold if it is not None.

Marei33:
test_mean_teacher.py should now be resolved

PR #401, opened 2024-11-03
Update vision mamba backbone to latest updates

anwai98:
This PR focuses on two things:
- [x] Adapt ViM-UNet's vim encoder to the latest legacy updates (the latest changes update some implementation details which in principle speed up the training time)

I decided to leave out the part of refactoring the installation step (in principle, I would like to move all installations in a config file which takes care of it all)


anwai98:
I did a sanity check training experiment on LIVECell, and I can reproduce numbers from our paper. This is now GTG from my side. Thanks!

PR #404, opened 2024-11-06
Add MOTUM dataset

anwai98:
This PR adds motum dataset, for tumor segmentation in brain MRIs. GTG from my side!

### DIFF
--- a/torch_em/data/datasets/medical/motum.py
+++ b/torch_em/data/datasets/medical/motum.py
@@ -19,7 +19,7 @@ from .. import util
 
 
 URL = "https://doi.gin.g-node.org/10.12751/g-node.tvzqc5/10.12751_g-node.tvzqc5.zip"
-CHECKSUM = ""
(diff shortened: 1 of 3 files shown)


torch_em/data/datasets/medical/motum.py
```diff
@@ -0,0 +1,141 @@
...
+
+from torch.utils.data import Dataset, DataLoader
+
+import torch_em
+
+from .. import util
+
+
+URL = "https://doi.gin.g-node.org/10.12751/g-node.tvzqc5/10.12751_g-node.tvzqc5.zip"
+CHECKSUM = ""
```
constantinpape, line 22: What about the checksum here? Does setting it to `""` actually work? I would assume that it fails because it has to be `None` if no checksum is provided?!

anwai98: I forgot to add that. I've updated it in the latest commit now. Thanks for spotting!

constantinpape (approved):

PR #408, opened 2024-11-09
Add TOIAM dataset

anwai98:
This PR adds TOIAM dataset for microbial live-cell segmentation (this data also has tracking information, would be super useful). GTG from my side!


constantinpape (approved):
Interesting dataset!

Issue #407, opened 2024-11-07
Add notebook for segmentation workflows


anwai98:
The idea is to add a notebook right from training segmentation models using custom loaders, and infer with and without tiling window based predictions (or with padding) and comment on the parameter choices (this would in turn hint towards the `torch-em`-supported functionalities as well)

I'll take care of this!

anwai98:
We have taken care of this now. I'll close this.

PR #409, opened 2024-11-12
Add notebook for tiling window-based prediction

anwai98:
This PR adds a notebook tutorial for tiling window based prediction supported by `torch-em`. GTG from my side!


anwai98:
Would be nice to add ReviewNB here (I checked from my side and I cannot add it here). I think @constantinpape would need to add this here via ReviewNB.

Resolves https://github.com/constantinpape/torch-em/issues/407.

PR #410, opened 2024-11-12
Add BlastoSPIM dataset

anwai98:
This PR adds a microscopy data for 3d nucleus segmentation in mouse embryos. GTG from my side!


constantinpape (approved):
Nice! And looks like this also contains annotations for tracking data.

PR #411, opened 2024-11-13
Add support to avoid overwriting trained models

anwai98:
This PR adds optional support to the `DefaultTrainer` in favor of avoiding to train if the training is already completed. This is controlled via the `fit` method with the `overwrite_training` argument, which is set to `True` as default (i.e. the current action of the trainers). If this is set to `False`, it will check in the `latest.pt` checkpoint to verify if the training is finished or not. GTG from my side!


anwai98:
Hi @constantinpape,

I took care of interactions between continuing training with overwriting trained models (i.e. raising error when `overwrite_training` is set to `False` and user passes a custom checkpoint to continue training)

GTG from my side!

PR #422, opened 2024-11-29
Add Vision Transformer from ScaleMAE

anwai98:
This PR adds the vision transformer integration from ScaleMAE and plugs it into the UNETR backbone (thanks to @Mareike79 for the implementation)

A few details to share with @Mareike79:
- In our previous setup, the dimension mismatch came from the issue that the outputs from the attention heads were flattened. They need to be restored to the target patch shape. This takes care of the issue we discussed.
- In addition, the model loading with your setup did not exactly work. Also, there were some parts of the pretrained model which are necessary to be dropped out (eg. decoder-related parameters, FPN and FCN heads, etc.)
- I added the configurations for `vit_b`, `vit_l` and `vit_h` (from what I understand, we have pretrained the model on `vit_b` and would be interested to observe the downstream task now)
- I disconnected the vision transformer class from the ScaleMAE repo because they do not have installation support for their module, which makes it difficult to fetch submodules from their original repo (eg. functions like `CustomCompose`, `get_2d_sincos_pos_embed_with_resolution`, etc).

PS. I'll leave this PR open for now. We work on this branch at the moment, and merge it later if everything works as desired.

PPS. This is still a work-in-progress.

cc: @constantinpape

### DIFF
--- a/torch_em/model/vit.py
+++ b/torch_em/model/vit.py
@@ -4,7 +4,7 @@ from functools import partial
 import torch
 import torch.nn as nn
 
-# we catch ImportErrors here because segment_anything, micro_sam and timm should
+# we catch ImportErrors here because segment_anything, micro_sam, scale_mae and timm should
 # only be optional dependencies for torch_em
 try:
     from segment_anything.modeling import ImageEncoderViT
@@ -14,10 +14,11 @@ except ImportError:
     _sam_import_success = False
 
 try:
-    from timm.models.vision_transformer import VisionTransformer
+    from timm.models.vision_transformer import VisionTransformer, PatchEmbed
     _timm_import_success = True
 except ImportError:
     VisionTransformer = object
+    PatchEmbed = object
     _timm_import_success = False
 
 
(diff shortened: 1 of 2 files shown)


anwai98:
Hi @constantinpape,

To keep things simple for future experiments, I will go ahead and merge this.

Everything works here, and all experiments ran by @Mareike79 are running via this branch!

I will go ahead and merge this. Let me know if you spot something here!

torch_em/model/vit.py
```diff
@@ -159,13 +156,220 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    def forward(self, x):
+        x, list_from_encoder = self.forward_features(x)
+        return x, list_from_encoder
+
+
+def get_vision_transformer(backbone: str, model: str, img_size: int = 1024, **kwargs) -> nn.Module:
     """Get vision transformer encoder.
 
     Args:
         backbone: The name of the vision transformer implementation. One of "sam" or "mae".
```
constantinpape, line 368: This should also list `scalemae` now.

constantinpape (commented):
I quickly checked it, looks good to me except for one docstring update.

torch_em/model/vit.py
anwai98: https://github.com/constantinpape/torch-em/pull/541 should sort this out as well. Thanks @constantinpape!

PR #438, opened 2024-12-14
Update documentation

anwai98:
This PR adds SynapseNet and original code repo for all other (known to my knowledge) projects using `torch-em`. I'll merge this once the tests pass!


constantinpape:
We could also add micro-sam

anwai98:
`micro-sam` is already in doc 😉 (both the preprint and original repo)

PR #440, opened 2024-12-16
Add BriFiSeg dataset

anwai98:
This PR adds a brightfield image dataset for nuclei segmentation. GTG from my side!

PS. It's a nice dataset but not of any immediate benefit to us from what I can think of


constantinpape (approved):

constantinpape:
Thanks! And that's quite relevant for one of Julia's current projects.

PR #444, opened 2024-12-18
Ad explicit splits for histopathology datasets

anwai98:
Superseeds and closes https://github.com/constantinpape/torch-em/pull/443.

This PR ensures that all datasets have exclusive train-val-test splits (thanks to @titusgriebel for taking care of this).

I am pending a few minor touch-ups. I'll merge it once I am done with them as well!


anwai98:
This is GTG from my side now. We should double check this if everything works as expected and there's no possible leakages for now!

cc: @titusgriebel

PR #446, opened 2024-12-19
Add CVZ Fluo dataset

anwai98:
This PR adds a new dataset for cell and nuclei segmentation in fluorescence microscopy images. GTG from my side!

PS: @constantinpape: This dataset is actually very good with very nice annotation tbh. I was too tempted to let this slide. Thanks for reminding. I will work on automating downloading via Synapse later (I found a tutorial on their documentation which seems (and is) super easy actually).

### DIFF
--- a/torch_em/data/datasets/light_microscopy/cvz_fluo.py
+++ b/torch_em/data/datasets/light_microscopy/cvz_fluo.py
@@ -1,7 +1,7 @@
 """The CVZ-Fluo dataset contains annotations for cell and nuclei segmentation in
 fluorescence microscopy images.
 
-NOTE: You need to install synapseclient`, create an account and authentication token on Synapse,
+NOTE: You need to install 'synapseclient', create an account and authentication token on Synapse,
 and follow the documentation to create your authentication config: https://python-docs.synapse.org/tutorials/authentication/
 and install using 'synapse get -r syn27624812'.
 
@@ -47,6 +47,23 @@ def get_cvz_fluo_data(path: Union[os.PathLike, str], download: bool = False):
         raise AssertionError(f"The dataset is not downloaded. Please download it manually from '{URL}'.")
 
 
+def _preprocess_labels(label_paths):
+    neu_label_paths = []
+    for lpath in tqdm(label_paths, desc="Preprocessing labels"):
+        neu_lpath = lpath.replace(".png", ".tif")
+        neu_label_paths.append(neu_lpath)
+        if os.path.exists(neu_lpath):
+            continue
+
+        if not os.path.exists(lpath):  # HACK: some paths have weird spacing nomenclature.
+            lpath = Path(lpath).parent / rf" {os.path.basename(lpath)}"
+
+        label = imageio.imread(lpath)
+        imageio.imwrite(neu_lpath, connected_components(label).astype(label.dtype), compression="zlib")
+
+    return neu_label_paths
+
+
 def get_cvz_fluo_paths(
     path: Union[os.PathLike, str],
     stain_choice: Literal["cell", "dapi"],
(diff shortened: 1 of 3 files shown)


torch_em/data/datasets/light_microscopy/cvz_fluo.py
```diff
@@ -0,0 +1,167 @@
+"""The CVZ-Fluo dataset contains annotations for cell and nuclei segmentation in
+fluorescence microscopy images.
+
+NOTE: You need to install 'synapseclient`, create an account and authentication token on Synapse,
```
constantinpape, line 4:
```suggestion
NOTE: You need to install synapseclient`, create an account and authentication token on Synapse,
```

torch_em/data/datasets/light_microscopy/cvz_fluo.py
```diff
@@ -0,0 +1,167 @@
...
+    """Obtain the CVZ-Fluo dataset.
+
+    Args:
+        path: Filepath to a folder where the downloaded data is saved.
+        download: Whether to download the data if it is not present.
+    """
+    if download:
+        raise NotImplementedError("Automatic download is not implemented for this dataset.")
+
+    data_dir = os.path.join(path, r"Annotation Panel Table.xlsx")
```
constantinpape, line 43: Why do you need to cast this to a raw string here?

torch_em/data/datasets/light_microscopy/cvz_fluo.py
```diff
@@ -0,0 +1,167 @@
...
+    if stain_choice not in ["cell", "dapi"]:
+        raise ValueError(f"'{stain_choice}' is not a valid stain choice.")
+
+    raw_paths = natsorted(
+        glob(os.path.join(path, data_choice, f"*-Crop_{stain_choice.title()}_Png.png"), recursive=True)
+    )
+    label_paths = [p.replace("_Png.png", "_Mask_Png.png") for p in raw_paths]
+
+    neu_label_paths = []
+    for lpath in tqdm(label_paths, desc="Preprocessing labels"):
```
constantinpape, line 85: I would refactor this in a small helper function, since you are processing the labels.

torch_em/data/datasets/light_microscopy/cvz_fluo.py
```diff
@@ -0,0 +1,167 @@
...
+        neu_lpath = lpath.replace(".png", ".tif")
+        neu_label_paths.append(neu_lpath)
+        if os.path.exists(neu_lpath):
+            continue
+
+        if not os.path.exists(lpath):  # HACK: some paths have weird spacing nomenclature.
+            lpath = Path(lpath).parent / rf" {os.path.basename(lpath)}"
```
constantinpape, line 92: Why do we need a raw string here?

torch_em/data/datasets/light_microscopy/cvz_fluo.py
```diff
@@ -0,0 +1,167 @@
...
+
+import imageio.v3 as imageio
+from skimage.measure import label as connected_components
+
+from torch.utils.data import Dataset, DataLoader
+
+import torch_em
+
+from .. import util
+from .neurips_cell_seg import to_rgb
```
constantinpape, line 27: How many channels does the data have and how exactly do we deal with it?

torch_em/data/datasets/light_microscopy/cvz_fluo.py
anwai98: With filenames having space between them, they create some issues on Windows. Scripting the paths with spaces as raw string usually avoids this issue!

torch_em/data/datasets/light_microscopy/cvz_fluo.py
anwai98: Okie

torch_em/data/datasets/light_microscopy/cvz_fluo.py
anwai98: There are a few labels which are named as `/path/to/ IMAGE.tif`, where there is a weird space infront of the capital "I". This creates issues and the path conventions cannot handle this apparently. So I just handle it by adding a space infront and allowing it as a raw string (due to the same reason as mentioned above)

torch_em/data/datasets/light_microscopy/cvz_fluo.py
anwai98: The ROIs for cell images have all 3-channels. However for ROIs with DAPI-staining, some images have 1 channel and others 3, which leads to collating issues.

torch_em/data/datasets/light_microscopy/cvz_fluo.py
anwai98: This is taken care of now. Thanks!

constantinpape (approved):

PR #456, opened 2024-12-25
Add splits for PENGWIN dataset

anwai98:
This PR adds data splits for PENGWIN datasets. It's still a WIP!


anwai98:
The X-Ray images here are synthetic (I wouldn't prefer to use them for our experiments). I'll discard the splitting plans for now!

PR #460, opened 2024-12-31
Minor update to three-channel logic in tissuenet

anwai98:
This PR updates the three-channel logic for TissueNet dataset. The new logic states the channels as follows: (cell channel, nucleus channel, channel with zeros). GTG from my side!


anwai98:
I'll go ahead and merge this @constantinpape. Let me know if there are modifications for which you have alternate suggestions! Thanks!

constantinpape (commented):
Looks good to me.

PR #463, opened 2025-01-02
Complete doc strings

constantinpape:


constantinpape:
Parts for which I will not yet finish the documentation:
- torch_em.model.probabilistic_unet
- torch_em.model.vim
- torch_em.self_training.loss.ProbabilisticUNetLoss
- torch_em.self_training.probabilistic_unet_trainer.py

The probabilistic U-Net functionality should be refactored since it is not only relevant for self-training.

Issue #466, opened 2025-01-04
Finalize the documentation and some refactoring


constantinpape:
I have updated most doc strings in #463. But the doc strings for the following are still missing:
- `torch_em.model.vim`: I am not sure how stable this is or if we should update it to newer (vision) mamba implementations.
- `torch_em.model.probabilistic_unet`: I am also not sure how finalized this is, given that we don't have support for 3D yet.
- `torch_em.self_training.loss.ProbabilisticUNetLoss`: Ths should be refactored, as this is more general than self-training.
- `torch_em.self_training.probabilistic_unet_trainer.py`: Should also be refactored.

PR #467, opened 2025-01-04
Update environment files and github workflows

constantinpape:


constantinpape:
@anwai98 this finalizes the documentation (except for #466 , which we can take care of later).

I have changed quite a few things in the dataset / dataloader description to improve the structure. 
If you have time you could give this a quick read to check if I missed something in these changes.

.github/workflows/build_docs.yaml
```diff
@@ -1,57 +1,58 @@
...
+    - name: Install pdoc
+      shell: bash -l {0}
+      run: pip install pdoc
 
-      - name: Install package
-        shell: bash -l {0}
-        run: pip install --no-deps -e .
+    - name: Generate Documentation
+      shell: bash -l {0}
+      run: pdoc synapse_net/ -d google -o doc/
```
anwai98, line 44: This should be `pdoc torch_em/ ...` right?

anwai98 (approved):
Thanks for documenting all functionalities in details. The dataloader / dataset doc and everything else looks good to me!

PS. I left some minor questions below.

environment.yaml
```diff
@@ -0,0 +1,16 @@
...
+    torch-em-dev
+dependencies:
+    - affogato
+    - bioimageio.spec
+    - bioimageio.core
+    - imagecodecs
+    - python-elf
+    - pytorch
+    - tensorboard
+    - tifffile >=2024
```
anwai98, line 13:
```suggestion
    - tifffile >=2024.9.20
```

constantinpape: Is there a reason for such a hard pin? I.e. is there some specific feature in this version that we need? Otherwise it's best practice not to pin versions too hard.

anwai98: Does it also work with versions structures such as `tifffile >=2024`? (the versions for `tifffile` are named as `tifffile: 2024.9.20`).

If `tifffile >=2024` works as expected, then that's what we want. Then we don't need a hard-pin.

constantinpape: Yes, this works just as in semantic versioning. `>= 2024` is valid for any version that starts with `2024, 2025 etc.`

anwai98: I see. Thanks for posting

PR #469, opened 2025-01-08
Add KiTS dataset

anwai98:
This PR adds KiTS dataset, for kidney and tumor segmentation in CT scans. It's still a WIP. I'll put it to review once I have verified the preprocessing!


anwai98:
I'll go ahead and merge this for now. This is a very relevant dataset for some other experiments planned for us!

PS. @constantinpape Let me know if you spot some logic which can be handled better!

PR #474, opened 2025-01-13
Add GlaS dataset

anwai98:
Superseeds and closes https://github.com/constantinpape/torch-em/pull/473.

This PR adds a new dataset for gland segmentation in histopathology (relevant for our current round of experiments).


anwai98:
@constantinpape GTG from my side!

And thanks to @titusgriebel for adding the dataset!

PR #475, opened 2025-01-15
fixed argument positions

titusgriebel:
@anwai98 argument positions of resize_inputs and download fixed


anwai98:
Good catch @titusgriebel! I'll merge this once the tests pass!

PR #493, opened 2025-03-04
 PUMA / CONIC semantic

titusgriebel:
@anwai98 
For CONIC:
- there is a pre-defined split that was generated with a [specific logic](https://github.com/vqdang/hover_net/blob/conic/generate_split.py). It creates 10 folds whereas the IDX 0 was used to train the HoVerNet baseline of the challenge (I would suggest sticking to it for our training and evaluation). This is implemented in a separate [patho-sam branch](https://github.com/computational-cell-analytics/patho-sam/blob/additional-semantic/experiments/data/semantic/create_conic_split.py). I was unsure whether to add this to the torch-em script. 
- the download logic currently does not take into account that download could be set to false without data being there already. Maybe you have an elegant way to implement this.

### DIFF
--- a/torch_em/data/datasets/histopathology/__init__.py
+++ b/torch_em/data/datasets/histopathology/__init__.py
@@ -1,4 +1,5 @@
 from .bcss import get_bcss_loader, get_bcss_dataset
+from .conic import get_conic_loader, get_conic_dataset
 from .consep import get_consep_loader, get_consep_dataset
 from .cpm import get_cpm_loader, get_cpm_dataset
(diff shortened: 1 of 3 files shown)


titusgriebel:
@anwai98 I updated the scripts as discussed!

anwai98:
Hi @titusgriebel,

Thanks for the PR. I will check it out in a bit. Meanwhile, would be nice to set up the downstream script for experiments already, so that once this is finalized and merged to `main`, we can fire up the GPUs! ;)

PR #503, opened 2025-03-30
Add CellMap dataset

anwai98:
WIP!


anwai98:
Okay, this PR is ready for it's first review!

I checked it locally for me, for one crop only. And it works as expected.

I will check back later in the evening if they have a crop-organelle mapping function to make my life easier hehe, otherwise I will script one (should be straight forward).

Let me know how it looks @constantinpape!

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+    
+    Inspired by https://github.com/janelia-cellmap/cellmap-segmentation-challenge/blob/main/src/cellmap_segmentation_challenge/cli/fetch_data.py
+
+    NOTE: The download scripts below are intended to stay as close to the original `fetch-data` CLI,
+    in order to ensure easy syncing with any changes to the original repository in future.
+    """  # noqa
+
+    # Importing packages locally.
+    # NOTE: Keeping the relevant imports here to avoid `torch-em` throwing missing module error.
+    import time
```
constantinpape, line 41: Can you move these imports to the top level? Especially for general packages like `time`, `h5py` and `xarray` that we have imported anyways.

For the other `cellmap` specific packages you can leave the import here.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+
+    from xarray import DataArray
+    from xarray_ome_ngff import read_multiscale_group
+    from xarray_ome_ngff.v04.multiscale import transforms_from_coords
+
+    from cellmap_segmentation_challenge.utils.fetch_data import read_group, subset_to_slice
+    from cellmap_segmentation_challenge.utils.crops import fetch_crop_manifest, get_test_crops
+
+    # Some important stuff.
+    padding = 0
```
constantinpape, line 54: Why do we hard-code this to 0? Shouldn't it be a parameter? (and I would then also not set the default to 0, but we can discuss a good default value later)

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+
+import torch_em
+
+from .. import util
+
+
+def _download_cellmap_data(
+    path: Union[os.PathLike, str],
+    organelles: Optional[Union[str, Sequence[str]]] = None,
+    crops: Sequence[str] = ("all"),
```
constantinpape, line 27: Your default is not a `Sequence`: (`"all"`) evaluates to `"all"`, i.e. this is a single string. And I think it's cleaner to have
```python
crops: Union[str, Sequence[str]] = "all"
```

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+        test_crops = get_test_crops()
+        log.info(f"Found '{len(test_crops)}' test crops.")
+
+    # Fetch all the crop manifests.
+    if crops == "all":
+        crops_parsed = crops_from_manifest + test_crops
+    elif crops == "test":
+        crops_parsed = test_crops
+    else:  # Otherwise, custom crops are parsed.
+        crops_split = tuple(int(x) for x in crops.split(","))
```
constantinpape, line 75: This is inconsistent with the expectation that you get a Sequence of strings, the code looks like you expect a single string with `","` separated ids. For the record, I think having a sequence of strings is cleaner.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+
+        crop_group_inventory = tuple(fn.split(store_path)[-1] for fn in gt_files)
+        crop_group_inventory = tuple(curr_cg[1:].split("/")[0] for curr_cg in crop_group_inventory)
+        crop_group_inventory = np.unique(crop_group_inventory).tolist()
+        crop_group_inventory = [curr_cg for curr_cg in crop_group_inventory if curr_cg not in [".zattrs", ".zgroup"]]
+
+        gt_crop_shape = gt_source_group[f"all/{resolution}"].shape  # since "all" exists "al"ways, we rely on it.
+
+        # Get the offset values for the ground truth crops.
+        for _, group in gt_source_group.groups():
```
constantinpape, line 128: Do we always get the lowest resolution group with this? It looks a bit dangerous to rely on this ...

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+        # Pad the slices (in voxel space)
+        slices_padded = tuple(
+            slice(max(0, sl.start - padding), min(sl.stop + padding, dim), sl.step)
+            for sl, dim in zip(slices, em_s0_array.shape)
+        )
+
+        # Extract cropped EM volume from remote zarr files.
+        em_crop = em_s0_array[tuple(slices_padded)].data.compute()
+
+        from concurrent.futures import ThreadPoolExecutor, as_completed
```
constantinpape, line 168: Move imports to the start of the file.

torch_em/data/datasets/electron_microscopy/cellmap.py
constantinpape: Ok, nvm, it looks like this is happening below (line 139). But the code here is still a bit unclear to read. Maybe another comment on what's going on would help.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+            # Store metadata
+            f.attrs["crop_id"] = crop.id
+            f.attrs["scale"] = scale.scale
+            f.attrs["translation"] = translation.translation
+
+            # Store inputs.
+            f.create_dataset(name="raw_crop", shape=em_crop.shape, data=em_crop, compression="gzip")
+
+            def _fetch_and_write_label(label_name):
+                gt_crop = gt_source_group[f"{label_name}/{resolution}"][:]
```
constantinpape, line 183: Would it be more efficient to use the dask array wrapper here as well?

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+    log.info(f"Done after {time.time() - fetch_save_start:0.3f}s")
+    log.info(f"Data saved to '{dest_path_abs}'.")
+
+    return path
+
+
+def get_cellmap_data(
+    path: Union[os.PathLike, str],
+    organelles: Optional[Union[str, Sequence[str]]] = None,
+    crops: Sequence[str] = ("all"),
```
constantinpape, line 208: See comment about `"crops"` and default values above.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+import numpy as np
+
+from torch.utils.data import Dataset, DataLoader
+
+import torch_em
+
+from .. import util
+
+
+def _download_cellmap_data(
```
constantinpape, line 24: Another general design comment: you typically don't want to repeat default arguments in private or helper functions. I.e., don't give defaults for any argument here and have the defaults only in the main argument. In this case `get_cellmap_data`. The reason for this is that you don't need to update the defaults in two places if they change.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -0,0 +1,339 @@
...
+    path: Union[os.PathLike, str],
+    organelles: Optional[Union[str, Sequence[str]]] = None,
```
constantinpape, line 26: This is currently not used. (You're probably aware, just to make sure.)

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Good point. I moved the common ones, and now added a try-catch to inform users about missing packages and pointing them on how to install `cellmap_segmentation_challenge` package. Thanks!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Ah yeah. I didn't notice that. I will make it an argument for now (also goes in the direction of our padding discussions).

PS. I won't resolve this comment yet for a self-reminder to come back to this!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Ahha yeah, that's better actually. Thanks!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: I will leave this as it is (as it comes from the cellmap challenge code - this will help us sync it with any new changes). For now, I plan to align all provided values to `crops` and convert them in `get_cellmap_data` in a structure that `csc` likes!

anwai98: Oh I just realized that `csc` is such a cool parallel name to `ctc`, but for segmentation challenge hehe!

PS. They do call their CLI with `csc` btw!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Good point. I will quickly check this!

anwai98: I just checked. For such small label crops, it doesn't make a difference :(

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Ahha okay. That's good to know. Thanks Constantin!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Yes, I am still thinking how to make it work. I'll check it out.

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: Ah and btw, I took care of this as I mentioned above.

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: I think I have a setup in place now. I make some explicit checks to validate whether organelle-crop combination matches or not.

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: I still need to think how to do this depending on the new observation I had today (i.e. crops can be of different shapes).

What would you recommend @constantinpape?

(the simplest I can think of is to pad all crops to the nearest (or next) multiple of 128 - i.e. (400, 400, 400) would become (512, 512, 512))

anwai98:
Okay, this PR is ready for it's next review now. Let me know how it looks @constantinpape!

torch_em/data/datasets/electron_microscopy/cellmap.py
anwai98: I added the padding scripts and kept the logic very simple (pad pixels along axis as requested by users)

constantinpape:
Looks good now. I made some tiny changes in the type annotations. I would expose the padding as a parameter (see also the comment), otherwise good to go from my end.

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -199,83 +223,122 @@ def _fetch_and_write_label(label_name):
...
-    _data_path = _download_cellmap_data(
-        path=data_path, organelles=organelles, crops=crops, resolution=resolution, download=download
+    _data_path, final_crops = _download_cellmap_data(
+        path=data_path,
+        crops=crops,
+        resolution=resolution,
+        download=download,
+        # NOTE: Set this to 0 for no padding or a higher number to pad in all dims.
+        # eg. with 16, it means that along each dims, we pad them with this value of pixels.
+        padding=16,  # NOTE: @CP: Should I expose this param / make it a global variable?
```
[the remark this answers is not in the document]
anwai98: And it's done!

anwai98:
Thank you so much @constantinpape! I've exposed `padding` as a parameter now. GTG from my side!

constantinpape (approved):

PR #505, opened 2025-04-03
Parallelization of prediction blocks

schilling40:
The prediction of the distance U-Net using a halo is parallelized using a list with block indexes as an input.

Functional implementation

### DIFF
--- a/torch_em/util/prediction.py
+++ b/torch_em/util/prediction.py
@@ -136,6 +136,7 @@ def predict_with_halo(
     tqdm_desc: str = "predict with halo",
     prediction_function: Optional[Callable] = None,
     roi: Optional[Tuple[slice]] = None,
+    iter_list: Optional[List[int]] = [],
 ) -> ArrayLike:
     """Run block-wise network prediction with a halo.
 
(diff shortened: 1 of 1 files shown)


torch_em/util/prediction.py
```diff
@@ -136,6 +136,7 @@ def predict_with_halo(
     tqdm_desc: str = "predict with halo",
     prediction_function: Optional[Callable] = None,
     roi: Optional[Tuple[slice]] = None,
+    iter_list: Optional[List[int]] = [],
```
constantinpape, line 139: A minor comment: `Optional` means that the value can be `None` and typically the default value should then also be `None` (the usual pattern in python).

torch_em/util/prediction.py
```diff
@@ -246,7 +247,11 @@ def predict_block(block_id):
                 output[bb] = prediction
 
     n_blocks = blocking.numberOfBlocks
+    if 0 == len(iter_list):
```
constantinpape, line 250: Better:
```python
if iter_list is None:
  ...
```

constantinpape (approved):

PR #510, opened 2025-04-11
Add BatchRenorm implementation

### LINKED ISSUE #499: Implement BatchRenorm feature normalization

anwai98:
WIP!

So it seems like we need a custom reimplementation (the `torchrl` one currently supports are limited to 1d).

I will check a quick 2d and 3d training to validate if it works, otherwise the implementation is kind of directly taken from https://pytorch.org/rl/main/_modules/torchrl/modules/models/batchrenorm.html#BatchRenorm1d wthout much changes at all.

Let me know how it looks @constantinpape


anwai98:
One of the first observations is that the training does not work with mixed precision (throws `NaN` after the first few iterations) (works fine without mixed precision though).

I will take a careful look on my implementation again!

constantinpape:
Do you know which implementation Alyona used?

anwai98:
> Do you know which implementation Alyona used?

Good point. I'll text her (but I am almost certain that she probably had to script it as well)

anwai98:
Ah btw, this works now! I will wait for Alyona to answer. I will double check the logic behind this once again over the weekend, otherwise should be working as expected imo! (we can quickly benchmark some stuff with this)

Issue #511, opened 2025-04-14
cellmap data download error


lufre1:
Tried to download the crops with the fucntion `get_cellmap_paths`
Up until now it worked fine, but
Crop with ID 557 throws following error:
```
2025-04-14 17:27:26 [info     ] The crop '378' is already saved at '/scratch-grete/projects/nim00007/data/cellmap/data_crops/crop_378.h5'. crop_id=378 dataset=jrc_zf-cardiac-1
2025-04-14 17:27:26 [info     ] The crop '379' is already saved at '/scratch-grete/projects/nim00007/data/cellmap/data_crops/crop_379.h5'. crop_id=379 dataset=jrc_zf-cardiac-1
2025-04-14 17:27:26 [info     ] The crop '380' is already saved at '/scratch-grete/projects/nim00007/data/cellmap/data_crops/crop_380.h5'. crop_id=380 dataset=jrc_zf-cardiac-1
2025-04-14 17:27:26 [info     ] The crop '381' is already saved at '/scratch-grete/projects/nim00007/data/cellmap/data_crops/crop_381.h5'. crop_id=381 dataset=jrc_zf-cardiac-1
crop.gt_source TestCropRow(id=557, dataset='jrc_ctl-id8-1', class_label='test', voxel_size=(1.74, 2.0, 2.0), translation=(23837.13, 7599.0, 35999.0), shape=(600, 600, 600))
Traceback (most recent call last):
  File "/mnt/vast-nhr/home/freckmann15/u12103/synapse/cellmap/download_cellmap_data.py", line 31, in <module>
    main(args)
  File "/mnt/vast-nhr/home/freckmann15/u12103/synapse/cellmap/download_cellmap_data.py", line 14, in main
    path = get_cellmap_paths(path=path, padding=0, download=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/vast-nhr/home/freckmann15/u12103/torch-em/torch_em/data/datasets/electron_microscopy/cellmap.py", line 363, in get_cellmap_paths
    data_path, crops = get_cellmap_data(
                       ^^^^^^^^^^^^^^^^^
  File "/mnt/vast-nhr/home/freckmann15/u12103/torch-em/torch_em/data/datasets/electron_microscopy/cellmap.py", line 293, in get_cellmap_data
    _data_path, final_crops = _download_cellmap_data(
[...]

PR #512, opened 2025-04-14
Minor fix to CellMap data - remove test crops

anwai98:
This PR closes https://github.com/constantinpape/torch-em/issues/511 by removing the test crops from the manifest, as they do not have ground truth masks, hence we do not want them for now.

GTG from my side!

### DIFF
--- a/torch_em/data/datasets/electron_microscopy/cellmap.py
+++ b/torch_em/data/datasets/electron_microscopy/cellmap.py
@@ -18,6 +18,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 import h5py
 import numpy as np
 import pandas as pd
+from xarray import DataArray
 
 from torch.utils.data import Dataset, DataLoader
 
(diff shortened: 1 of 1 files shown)


lufre1:
> This PR closes #511 by removing the test crops from the manifest, as they do not have ground truth masks, hence we do not want them for now.
> 
> GTG from my side!

Thank you very much @anwai98 :blush:

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -65,15 +65,12 @@ def _download_cellmap_data(path, crops, resolution, padding, download=False):
...
-        test_crops = get_test_crops()
-        log.info(f"Found '{len(test_crops)}' test crops.")
-
-    # Fetch all the crop manifests.
     if crops == "all":
-        crops_parsed = crops_from_manifest + test_crops
+        crops_parsed = crops_from_manifest
     elif crops == "test":
-        crops_parsed = test_crops
+        raise NotImplementedError("The test crops do not have GT. Hence we do not allow using them in our setup.")
```
constantinpape, line 71: Should we address this right away, and just not process the ground-truth in this case? I would hope that this is relatively straight forward.

anwai98: You mean store only the EM volumes for these test crops?

constantinpape: Yes

anwai98: Okay, it's taken care of now ;)

anwai98:
This PR now also addresses getting the test crops, without labels. They are not considered atm for loading them in the dataloader, but if desired, the filepaths can be extracted from `get_cellmap_paths` by setting the argument: `return_test_crops` to `True`.

PR #517, opened 2025-04-24
Add PNAS Arabidopsis data

anwai98:
WIP!


anwai98:
And this is ready as well. One important thing here is to double check the publication. I think I found the right one, but I just wanna make sure.

Besides that, GTG from my side!

constantinpape (approved):

constantinpape:
> And this is ready as well. One important thing here is to double check the publication. I think I found the right one, but I just wanna make sure.

Yep, it's the correct one.

Issue #421, opened 2024-11-25
Wrap SAMed2D dataset to h5 format per sub-dataset


anwai98:
Instead of having the data scattered out in tif files (hurting the file count at the moment), would make sense to have files stored in container formats (preferably in hdf5 file format, else I would come back to the inodes problem again with zarr for instance).

I'll take care of this later once I have time!

PR #519, opened 2025-04-24
Start packing SA-Med2d data into container file-format

anwai98:
WIP!

I figured out a way for partial unzipping by looking for files / folders using search strings and only unzipping those files, eg. `unzip data.zip "*AMOS2022*"`

I will come back to this later once I catch up with the `medico-sam` stuff!

Once this works, this would resolve https://github.com/constantinpape/torch-em/issues/421.

### DIFF
--- a/scripts/datasets/medical/check_sa_med2d.py
+++ b/scripts/datasets/medical/check_sa_med2d.py
@@ -1,13 +1,19 @@
+import os
+import sys
+
 from torch_em.util.debug import check_loader
 from torch_em.data.datasets.medical import get_sa_med2d_loader
 
 
-ROOT = "/scratch/share/cidas/cca/data/sa-med2d"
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,8 @@ logs/
 *.pyc
 *.egg-info/
 *.n5/
+*.ipyng_checkpoints
```
constantinpape, line 7: Should this be `.ipynb_checkpoints`?

constantinpape (commented):
I looked through it, and the logic makes sense to me. I only found one typo (see comment). Please go ahead and merge it when it's ready from your side.

.gitignore
anwai98: Ahha yeah you are right. I will fix it right away!

PR #525, opened 2025-05-11
Add aiSEGcell data

anwai98:
WIP!

### DIFF
--- a/scripts/datasets/light_microscopy/check_aisegcell.py
+++ b/scripts/datasets/light_microscopy/check_aisegcell.py
@@ -1,5 +1,27 @@
+import os
+import sys
+
+from torch_em.util.debug import check_loader
+from torch_em.data.datasets import get_aisegcell_loader
+
+
+sys.path.append("..")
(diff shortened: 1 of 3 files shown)


anwai98:
This PR adds aiSEGcell data, for nucleus segmentation in (paired) BF and fluo images. The cool thing about this data is that it has ca. 20k labeled images!!

cc: @constantinpape

constantinpape:
It's nice to integrate this, but be aware that this is annotating the nucleus in brightfield images - we should not use it directly for training cell segmentation.

anwai98:
> this is annotating the nucleus in brightfield images - we should not use it directly for training cell segmentation.

They have fluo-labeled images as well, with which nuclei has been annotated. I thought that might help us for the nucleus segmentation part!

constantinpape:
> They have fluo-labeled images as well, with which nuclei has been annotated. I thought that might help us for the nucleus segmentation part!

Yes, that is a good point.

PR #536, opened 2025-06-11
Feature: added optional reproducible splitting logic for LyNSec dataset

titusgriebel:
@anwai98 I added an optional (so one can still use the entire dataset for sole evaluation like we did in the paper) splitting logic that follows the other datasets' logic.

For IHC TMA, I realized there is a curator-defined split already.


anwai98:
I merged this to `main` now. Please double check if the patches I made work as expected or there's anything I missed!

PR #540, opened 2025-06-18
Minor updates to fetch valid volume paths for respective organelles

anwai98:
I realized there's a easy way to filter out crops which do not have some organelles. Went ahead and implemented it. GTG from my side!

cc: @constantinpape @lufre1


anwai98:
Now that I think about it, I might add a resolution filtering on top of this (would make accessing images for a particular resolution sooo much easier). I'll take care of this later in the evening!

anwai98:
As I proposed yesterday, I added another strategy to filter crops via `voxel_size`. I think it's valuable to add this and would allow users requesting for specific crops with a desired resolution (eg. `[8.0, 8., 8.0]` being the most commonly available voxel spacing across most volumes).

Good to review from my side!

torch_em/data/datasets/electron_microscopy/cellmap.py
```diff
@@ -423,12 +430,40 @@ def get_cellmap_paths(
...
+    if organelles is None:
+        organelles = "all"
+
+    if isinstance(organelles, str):
+        organelles = [organelles]
+
+    other_volume_paths = []
+    for organelle in organelles:
+        for vpath in volume_paths:
+            if f"label_crop/{organelle}" in open_file(vpath) and vpath not in other_volume_paths:
```
lufre1, line 443: This checks only for a organelle key in the crops, there is one minor and one major issue:
1. (minor) e.g. "mito" and "mito_lum" are not always both present (sometimes just one is present) and this runs through the entire dataset.
2. (major) Label datasets can be empty! Or they are not present as a dataset but only in "label_crops/all" (via class ID)

constantinpape (approved):
Looks good from my end.

PR #548, opened 2025-09-09
Make unet implementation scriptable

FynnBe:
I started looking into improving affable-shark and its onnx weights in particular, which led me down the rabbit hole, so this PR makes the unet implementation mostly scriptable (with the small exception of the 'check_shape' validation). 

In order to make it scriptable I 
- added type annotations (in particular a few `Final` annotations to inform torch.jit about constants).
  Not all annotations are required, many I added for myself for a better overview and to help avoid unintended code changes.
- removed some of the`return_outputs` flags, changing the `Encoder` and `Decoder` interfaces slightly for static signatures needed for scripting.

I am ~~not~~ done testing the code with the affable-shark model example.


FynnBe:
I will try to allow `scale_factor` to be `List[float]` again, this seems to be used outside of unet.py...

PR #556, opened 2025-09-26
Minor update to MinInstanceSampler for reject ids

anwai98:
This PR makes tiny changes to `MinInstanceSampler`, to allow ids which can now be rejected / not considered in the sampling criterion! (and keeps the default behavior as before)

GTG from my side!

### DIFF
--- a/torch_em/data/sampler.py
+++ b/torch_em/data/sampler.py
@@ -122,11 +122,24 @@ class MinInstanceSampler:
         min_num_instances: The minimum number of instances for accepting a sample.
         p_reject: The probability for rejecting a sample that does not meet the criterion.
         min_size: The minimal size for instances to be taken into account.
+        reject_ids: The ids to reject (i.e. not consider) for sampling a valid input.
     """
-    def __init__(self, min_num_instances: int = 2, p_reject: float = 1.0, min_size: Optional[int] = None):
+    def __init__(
+        self,
+        min_num_instances: int = 2,
+        p_reject: float = 1.0,
+        min_size: Optional[int] = None,
+        reject_ids: Optional[List[float]] = None,
+    ):
         self.min_num_instances = min_num_instances
         self.p_reject = p_reject
         self.min_size = min_size
+
+        if reject_ids is None:
+            self.reject_ids = reject_ids
+        else:
+            assert isinstance(reject_ids, list)
+            self.reject_ids = [float(idx) for idx in reject_ids]
 
     def __call__(self, x: np.ndarray, y: np.ndarray) -> bool:
         """Check the sample.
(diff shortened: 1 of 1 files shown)


torch_em/data/sampler.py
```diff
@@ -122,12 +122,25 @@ class MinInstanceSampler:
...
         min_size: The minimal size for instances to be taken into account.
+        reject_ids: The ids to reject (i.e. not consider) for sampling a valid input.
     """
-    def __init__(self, min_num_instances: int = 2, p_reject: float = 1.0, min_size: Optional[int] = None):
+    def __init__(
+        self,
+        min_num_instances: int = 2,
+        p_reject: float = 1.0,
+        min_size: Optional[int] = None,
+        reject_ids: Optional[List[float]] = None,
```
constantinpape, line 132: The type of `id`s should be `int`, not `float`.

torch_em/data/sampler.py
```diff
@@ -122,12 +122,25 @@ class MinInstanceSampler:
...
+    ):
         self.min_num_instances = min_num_instances
         self.p_reject = p_reject
         self.min_size = min_size
 
+        if reject_ids is None:
+            self.reject_ids = reject_ids
+        else:
+            assert isinstance(reject_ids, list)
+            self.reject_ids = [float(idx) for idx in reject_ids]
```
constantinpape, line 142: Not sure why you would cast ids to float.

torch_em/data/sampler.py
```diff
@@ -139,10 +152,14 @@ def __call__(self, x: np.ndarray, y: np.ndarray) -> bool:
             Whether to accept this sample.
         """
         uniques, sizes = np.unique(y, return_counts=True)
+
         if self.min_size is not None:
             filter_ids = uniques[sizes >= self.min_size]
             uniques = filter_ids
 
+        if self.reject_ids is not None:
+            uniques = [idx for idx in uniques if idx not in self.reject_ids]
```
constantinpape, line 161: Is this really what you want? If you call this reject ids I would suspect that you want to actually reject a sample that contains them.

torch_em/data/sampler.py
anwai98: Some images have labels as floating points. But, it would probably be easier to assume integers in labels and treat them as that! (because we parse over unique ids). I'll update this!

torch_em/data/sampler.py
anwai98: Yes, that's the idea! (this is done to ignore background labels as a valid criterion) (eg. it's -1 for PlantSeg Ovules and 1 for PlantSeg Root, that are labeled as bakground), then once the sampler can consider a criterion without these ids, I can filter them out later via `label_transform`)

anwai98: So the application looks like: `sampler=MinInstanceSampelr(reject_ids=[-1])`

anwai98: Makes sense. I updated the argument in a new commit1

constantinpape (approved):

PR #560, opened 2025-10-06
Add cytoDArk0 dataset

anwai98:
This PR adds a new histopathology dataset (Nissl-stained) for cell instance segmentation! (I couldn't find intracellular classification labels, but we have different species, maybe we are interested in classifying between them?)

I'll shoot an email to the corresponding authors regarding this anyways. For the current state, GTM from my side!

might be relevant to cc: @titusgriebel and/or @caroteu


anwai98:
Okay, I got a response from the corresponding author, and they haven't and won't release the neuron vs glia cell classification labels right now. It'll be a part of their next project!

Let's go ahead and merge this then! (it's sufficient for our instance segmentation projects) @constantinpape

PR #569, opened 2025-12-16
Add 3d UNETR model

anwai98:
This PR adds the 3d UNETR model (i.e. the stacking encoder features with 3d convolutions / strip convolutions for the depth dimension as a choice in the decoder).

GTG from my side!

cc: @constantinpape @Marei33

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install segment-anything
         shell: bash -l {0}
-        run: mamba install -c conda-forge segment-anything
+        run: micromamba install -c conda-forge segment-anything
 
       - name: Install package
         shell: bash -l {0}


torch_em/model/unetr.py
```diff
@@ -408,11 +410,337 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        super().__init__()
+
+        self.img_size = img_size
+
+        # Step 1: the image encoder backbone.
+        image_encoder = get_vision_transformer(backbone=backbone, model=model_type, img_size=img_size)
+
+        # Load the pretrained image encoder weights
+        if checkpoint_path is not None:
+            state = torch.load(checkpoint_path, map_location="cpu")["model"]
```
constantinpape: I wouldn't hard-code the save key as `"model"`. I think that we need a single logic as https://github.com/constantinpape/torch-em/blob/main/torch_em/model/unetr.py#L49 to handle this for different models.

constantinpape (approved):
I went ahead and added a few minor changes so that the defaults actually work and added a test. This looks good from my end now and can be merged once the tests pass.

PR #575, opened 2026-01-07
added grid shift for the nifty blocking 

lufre1:
@constantinpape @anwai98 
by padding the input dynamically before creating the grid; unpadding subsequently
The grid is anchored at input[0, 0, 0] so left padding will move the entire grid without disrupting the existing implementation too much.
left padding can be adjusted per dimension based on the initial block shape:
e.g. block shape [32, 32, 32] with grid_shift [0, 0.25, 0.25] -> left_padding_voxels = [0, 8, 8]

### DIFF
--- a/torch_em/util/prediction.py
+++ b/torch_em/util/prediction.py
@@ -264,11 +264,13 @@ def predict_with_halo(
             if preprocess is not None:
                 inp = preprocess(inp)
 
+            # add (channel) and batch axis
             expand_dims = np.s_[None] if with_channels else np.s_[None, None]
             inp = torch.from_numpy(inp[expand_dims]).to(device)
 
             prediction = net(inp) if prediction_function is None else prediction_function(net, inp)
 
+            # allow for list of tensors
             try:
                 prediction = prediction.cpu().numpy().squeeze(0)
             except AttributeError:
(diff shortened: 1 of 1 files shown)


torch_em/util/prediction.py
```diff
@@ -81,16 +100,13 @@ def _load_block(input_, offset, block_shape, halo, padding_mode="reflect", with_
     starts = [off - ha for off, ha in zip(offset, halo)]
     stops = [off + bs + ha for off, bs, ha in zip(offset, block_shape, halo)]
 
-    # we pad the input volume if necessary
```
constantinpape: Please don't remove comments in un-related parts of the code.

constantinpape (commented):
I think that the implementation makes sense, but removal of unrelated comments makes the diff hard to read and is not a good practice in general. Also some variable definitions in the implementation seem redundant, see comments.

torch_em/util/prediction.py
```diff
@@ -227,31 +279,41 @@ def predict_block(block_id):
...
-                prediction[~mask_block] = 0
+                    mb = np.broadcast_to(mask_block[None], prediction.shape)
+                else:
+                    mb = mask_block
+                prediction = prediction.copy()
+                prediction[~mb] = 0
 
             bb = tuple(slice(beg, end) for beg, end in zip(block.begin, block.end))
-            if isinstance(output, list):  # we have multiple outputs and split the prediction channels
+            if isinstance(output, list):
```
constantinpape: Comment removed -- please restore

PR #578, opened 2026-01-11
Make embed_dim argument choice flexible for 3d UNETR

anwai98:
This PR fixes a bug in 3d UNETR, for adapting to all ViT backbones! (previously only worked for SAM2, my fault in hardcoding the `embed_dim` tbh)

Thanks to @Marei33 for reporting this! <3

Will merge this once the tests pass!


anwai98:
This shouldn't break anything functionally too, just a simple (yet critical) patch fix to make ViTs more flexible (the same as 2d UNETR)

cc: @constantinpape

PR #579, opened 2026-01-11
Edit UNETR3D

Marei33:
in `UNETR3d` changed `self.end_up` to `self.deconv_out` to stay consistent with the `UNETR2D` implementation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ logs/
 *.png
 *.out
 *.sh
+*.pth
+*.pt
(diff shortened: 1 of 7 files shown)


constantinpape (approved):

constantinpape:
> I am only worried about my currently pretrained models for `micro-sam2`, but I am at a stage where I will retrain models anyways.

I think it's import to be consistent here, so that UNETR2d and 3d can be used as interchangeably as possible. We don't have many finetuned SAM2 models yet and it should still be possible to load them with some custom code. So let's merge this.

PR #596, opened 2026-02-05
Add MitoEMv2 dataset

anwai98:
This PR adds the entire MitoEMv2 dataset (I was too excited about having access to the data already hehe, hence went ahead and incorporated it here). GTG from my side!

PS. The annotations are indeed phenomenal tbh.


anwai98:
I'll go ahead and merge this. On a second thought, I think we can merge v1 + v2 scripts in one place. But let's not do this until it's published (and maybe we wanna switch to BIA's OME-Zarrs which Peng is working on)

PR #603, opened 2026-02-06
Add YeastSAM data

anwai98:
This PR adds yeast cell segmentation data for DIC microscopy images. DIC is cool to have in general because I feel we need to improve in this modality in particular. I'll go ahead and merge this!


anwai98:
Oh, and YeastSAM is kind of build on `micro-sam` hehe. I wanted to check out the data for a while, it's quite nice tbh.

PR #643, opened 2026-03-18
raw dataset with masks

stmartineau99:
new subclass `RawDatasetWithMasks` handles:
- sample mask for sampling patches; intended for use with `MinForegroundSampler`
- background mask for subtracting background from the loss

changes
- new args `sample_mask`, `sample_mask_key`, `bg_mask`, `bg_mask_key`
- `_get_sample` - returns raw and a bb, which are needed to ensure the correct patch is extracted from `bg_mask`, if present
-  `_get_item` - if `bg_mask` is present, returns a stacked array (expected format) for `MeanTeacherTrainerWithBackgroundMask`

to do
- update documentation
- write tests

### DIFF
--- a/test/data/test_raw_dataset.py
+++ b/test/data/test_raw_dataset.py
@@ -133,6 +133,8 @@ class TestRawDatasetWithMasks(TestRawDatasetBase, unittest.TestCase):
             self.assertEqual(patch.shape, expected_shape)
 
             # check that patches are sampled inside the ROI defined by sample mask
+            # note that the raw data and the sample mask are identical, so this checks if the
+            # positive samples from the sample mask are correct
             self.assertGreaterEqual(patch.mean().item(), min_fraction)
 
     def test_bg_mask(self):
--- a/test/data/test_raw_dataset.py
+++ b/test/data/test_raw_dataset.py
@@ -4,9 +4,14 @@ import unittest
 import h5py
 import numpy as np
 
+from torch_em.data.raw_dataset import RawDataset, RawDatasetWithMasks
 
-class TestRawDataset(unittest.TestCase):
+
+class TestRawDatasetBase:
+    """Base class to hold shared tests for RawDataset and subclasses."""
+
     path = "./data.h5"
+    dataset_class = None
 
     def tearDown(self):
         os.remove(self.path)
@@ -19,11 +24,10 @@ class TestRawDataset(unittest.TestCase):
         return shape, chunks
 
     def test_dataset_3d(self):
-        from torch_em.data import RawDataset
         raw_key = "raw"
         shape, chunks = self.create_default_data(raw_key)
         patch_shape = chunks
-        ds = RawDataset(self.path, raw_key, patch_shape=patch_shape)
+        ds = self.dataset_class(self.path, raw_key, patch_shape=patch_shape)
         self.assertEqual(ds.raw.shape, shape)
         self.assertEqual(ds._ndim, 3)
         expected_shape = (1,) + patch_shape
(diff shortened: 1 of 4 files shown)


stmartineau99 (commented):
- updated `load_data` to handle mrc files 
- finished responding to the comments, then wrote tests for sample mask and bg_mask
- made a `TestRawDatasetBase` to store shared tests, these get reused for testing both classes

test/data/test_raw_dataset.py
```diff
@@ -19,66 +24,148 @@ def create_default_data(self, key, shape=None, chunks=None):
...
+        self.assertEqual(ds.sample_mask.shape, shape)
+        self.assertEqual(ds._ndim, 3)
+
+        expected_shape = (1,) + patch_shape
+        for i in range(10):
+            patch = ds[i]
+            self.assertEqual(patch.shape, expected_shape)
+
+            # check that patches are sampled inside the ROI defined by sample mask
+            self.assertGreaterEqual(patch.mean().item(), min_fraction)
```
constantinpape, line 136: I am not sure if this check makes sense. Your criterion is defined w.r.t. the `sample_mask`, but you evaluate it w.r.t. the underlying data here.

torch_em/data/raw_dataset.py
```diff
@@ -199,3 +199,184 @@ def __setstate__(self, state):
...
         self.__dict__.update(state)
+
+
+class RawDatasetWithMasks(RawDataset):
+    """Extends `RawDataset` to support a sample mask and a background mask.
+
+        - The sample mask is used by the sampler to extract patches from a region of interest, e.g.,
+            using `MinForegroundSampler`, to avoid empty patches. 
+        - The background mask is a binary mask identifying regions or structures that belong to the background. 
+            It can be used during unsupervised training to subtract background regions from the predicted
```
constantinpape, line 210: It would be good to describe more explicitly what happens:
"It will be appended as an additional channel to the returned data"
"It can then for example be used as an ignore label in unsupervised training."

constantinpape (commented):
Looks very good. I only found a potential issue with the test logic.

test/data/test_raw_dataset.py
stmartineau99: My intention was to check that patches are sampled inside the ROI defined by the sample mask, which mimics a lamella, where the central slices are foreground and above and below is background. Patches of shape (32, 32, 32) are sampled with the `MinForegroundSampler` sampler and then `self.assertGreaterEqual(patch.mean().item(), min_fraction)` checks if it has above 95% pixels originating from inside the ROI. Only raw is returned from the dataset, because the `sample_mask` is used by the sampler and then discarded. I therefore made the raw data the same as the sample mask data, so that this check works (pixels inside the ROI will be 1). I hope that makes sense.

stmartineau99 (commented):
I responded to your comment about the test.

test/data/test_raw_dataset.py
constantinpape: This makes sense. I overlooked before that the data is identical.

constantinpape (approved):

test/data/test_raw_dataset.py
```diff
@@ -19,66 +24,148 @@ def create_default_data(self, key, shape=None, chunks=None):
...
+        self.assertEqual(ds.raw.shape, shape)
+        self.assertEqual(ds.sample_mask.shape, shape)
+        self.assertEqual(ds._ndim, 3)
+
+        expected_shape = (1,) + patch_shape
+        for i in range(10):
+            patch = ds[i]
+            self.assertEqual(patch.shape, expected_shape)
+
+            # check that patches are sampled inside the ROI defined by sample mask
```
constantinpape, line 135:
```suggestion
            # check that patches are sampled inside the ROI defined by sample mask
            # note that the raw data and the sample mask are identical, so this checks if the
            # positive samples from the sample mask are correct
```

PR #650, opened 2026-03-23
Cldice loss and metric (new PR)

stmartineau99:
**remaking this PR form the branch torch_em/cldice_loss instead of my own fork**

clDice code copied from https://github.com/jocpae/clDice/tree/master

clDice metric

simple skeletonization (skimage.morphology)
Note - membrain-seg uses a much more complex skeletonization for surface dice https://github.com/teamtomo/membrain-seg/blob/main/src/membrain_seg/segmentation/skeletonize.py
I guess that their method performs better for membranes, I figured its worth looking at again for comparison
clDice loss

authors introduce soft skeletonization, which makes the loss fn differentiable


test/.DS_Store
constantinpape, line 1: Please remove this. (I will also update the .gitignore on main to avoid adding the mac specific files.)

torch_em/.DS_Store
constantinpape, line 1: Please remove.

.DS_Store
constantinpape, line 1: Please remove.

constantinpape (commented):
Looks good overall. Please remove the `.DS_Store` files. Then, we can merge it once checking the first results from your trainings.

constantinpape (approved):

stmartineau99 (commented):
I think this can be merged.

constantinpape (approved):

PR #654, opened 2026-04-10
Fix more typo issues in docstrings

anwai98:
Found some inconsistent docstrings here and there. Will fix them and merge this then!


anwai98:
The number of files touched up is a lot, but the changes are mostly super minor, either fixing type hints, docstrings, or some argument declarations here and there. I'll go ahead and merge this once the tests pass!

PR #656, opened 2026-04-12
Minor update to cremi dataset

anwai98:
Hi @constantinpape,

Okay this PR is a bit of a tricky one, but lemme explain:

When `defect_augmentation_kwargs` are used (which is necessary for CREMI to dodge data-level artifacts), it utilizes the `raw_transform` inside the `get_cremi_dataset` and ignores `raw_transform` as input from user completely (which is exactly how it should be). Now, I would like to keep the data statistics to 8 bit (i.e. standardized by `torch_em.transform.get_raw_transform`). For this, I allow users to pass `raw_transform` too, such that it acts as a normalization protocol on top of `defect_augmentation_kwargs`, when applied.

What do you think about this change?

### DIFF
--- a/torch_em/data/datasets/electron_microscopy/cremi.py
+++ b/torch_em/data/datasets/electron_microscopy/cremi.py
@@ -144,7 +144,9 @@ def get_cremi_dataset(
 
     # defect augmentations
     if defect_augmentation_kwargs is not None:
+        from torch_em.transform.raw import standardize
         raw_transform = torch_em.transform.get_raw_transform(
+            normalizer=kwargs.pop("raw_transform", standardize),
             augmentation1=torch_em.transform.EMDefectAugmentation(**defect_augmentation_kwargs)
         )
         kwargs = util.update_kwargs(kwargs, "raw_transform", raw_transform)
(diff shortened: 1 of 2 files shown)


constantinpape:
I am not sure if the defect augmentations are all well defined if we don't use `standardize` -> we can discuss together later.

anwai98:
Okie!

anwai98:
Hi @constantinpape,

What do you think about this now?

torch_em/data/datasets/electron_microscopy/cremi.py
```diff
@@ -144,7 +146,13 @@ def get_cremi_dataset(
 
     # defect augmentations
     if defect_augmentation_kwargs is not None:
+        if "raw_transform" in kwargs:
+            warnings.warn(
+                "'raw_transform' was found in kwargs. It will be used as the "
+                "normalizer for the defect augmentation pipeline."
```
constantinpape:
```suggestion
                "normalizer for the defect augmentation pipeline, which may lead to incorrect results"
                "if the normalizer maps to an unexpected data range."
```

constantinpape (approved):

PR #660, opened 2026-04-14
Add SegPath dataset

anwai98:
This PR adds a new H&E dataset with both cell-level and tissue-level semantic annotations. I'll make a few quick tests and then merge it!


anwai98:
Okay this is good to fire up. Just fyi: this seems like quite a lot of awesome data for us @constantinpape @titusgriebel 

I'll go ahead and merge this!

PR #663, opened 2026-04-16
Add UNETR input range checks before normalization

anwai98:
Hi @constantinpape,

Going along our discussion from this morning, what do you think about this defensive check before normalization for expected ranges?

### DIFF
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -319,6 +319,26 @@ class UNETRBase(nn.Module):
         pixel_std = torch.tensor(std, device=device, dtype=dtype).view(*view_shape)
         return pixel_mean, pixel_std
 
+    def _check_input_normalization_range(self, x: torch.Tensor, expected_range: Optional[Tuple[float, float]]) -> None:
+        """Check whether raw inputs match the value range expected by the model normalizer."""
+        if expected_range is None:
+            return
+
+        if not torch.all(torch.isfinite(x)):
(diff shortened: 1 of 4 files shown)


torch_em/model/unetr.py
```diff
@@ -319,6 +319,26 @@ def _as_stats(self, mean, std, device, dtype, is_3d: bool):
...
+        """Check whether raw inputs match the value range expected by the model normalizer."""
+        if expected_range is None:
+            return
+
+        if not torch.all(torch.isfinite(x)):
+            raise ValueError("The UNETR input contains NaN or infinite values before normalization.")
+
+        min_value, max_value = expected_range
+        is_out_of_range = torch.any((x < min_value) | (x > max_value))
+        looks_unit_scaled = x.is_floating_point() and max_value > 1.0 and torch.amax(x) <= 1.0
```
constantinpape, line 332: I don't quite get this check.

constantinpape (commented):
The overall logic looks good, but I don't quite understand the `looks_unit_scaled` check.

torch_em/model/unetr.py
anwai98: Good point actually (I was just trying to be conservative with the checks). I've simplified it by removing the strange `looks_unit_scaled` heuristic. Now we do a classic (finite value) range check.

anwai98:
Okay this should be ready now @constantinpape

constantinpape (approved):

PR #664, opened 2026-04-17
add invertible augmentations

Marei33:
- add FixMatchTrainerWithInvertibleAugmentations and MeanTeacherTrainerWithInvertibleAugmentations
- add loss functions SelfTrainingLossWithInvertibleAugmentations and SelfTrainingLossAndMetricWithInvertibleAugmentations to be compatible with different input logic
- add InvertibleAugmenter class + default classes for FixMatch and Mean Teacher augmenters
- update SelfTrainingTensorboardLogger to display all output channels


Marei33:
@constantinpape ready to review!

constantinpape (commented):
This looks very good from my end, but I think we should at least add a test for the invertible augmentations (transform and transform back for a pure geometric augmentation and make sure that it's roughly the identity).

constantinpape:
@anwai98 : can you also take a quick look here? The idea is to add new classes that supports Marei's invertible augmentations.

anwai98:
Thanks, I'll check this out in some!

torch_em/self_training/fix_match.py
```diff
@@ -361,3 +366,318 @@ def _validate_impl(self, forward_context):
...
+
+            with forward_context(), torch.no_grad():
+                # Compute the pseudo labels.
+                pseudo_labels, label_filter = self.pseudo_labeler(self.model, teacher_input)
+                pseudo_labels_inv = self.augmenter.teacher.reverse_transform(pseudo_labels)
+                label_filter_inv = self.augmenter.teacher.reverse_transform(label_filter)
+
+            pseudo_labels = pseudo_labels.detach()
+            if label_filter is not None:
+                label_filter = label_filter.detach()
```
anwai98, line 489: Do I understand correctly that `label_filter` can be `None` here (line 489)? If so, line 486 would break, because `reverse_transform` is called unconditionally on `label_filter`. I think this should be guarded, eg. `label_filter_inv = None if label_filter is None else self.augmenter.teacher.reverse_transform(label_filter)`

torch_em/self_training/fix_match.py
```diff
@@ -361,3 +366,318 @@ def _validate_impl(self, forward_context):
...
+
+            # Perform distribution alignment for pseudo labels
+            pseudo_labels = self.get_distribution_alignment(pseudo_labels)
```
anwai98, line 492: I think distribution alignment is ineffective in this invertible FixMatch path. `pseudo_labels_inv` is computed before `get_distribution_alignment(pseudo_labels)`, but the loss uses `pseudo_labels_inv`, not the aligned labels. So the alignment step does not affect training. Should the alignment be applied to `pseudo_labels` first, and only then inverse-transform the aligned pseudo-labels for the loss?

torch_em/self_training/fix_match.py
```diff
@@ -361,3 +366,318 @@ def _validate_impl(self, forward_context):
...
```
anwai98, line 563: Same comment as line 493 (see above)

And another similar occurrence later!

torch_em/self_training/fix_match.py
```diff
@@ -361,3 +366,318 @@ def _validate_impl(self, forward_context):
...
+                # We pass the model, the input and the labels to the supervised loss function,
+                # so that how the loss is calculated stays flexible, e.g. to enable ELBO for PUNet.
+                supervised_pred = self.model(xs)
+                supervised_loss = self.supervised_loss(supervised_pred, ys)
+
+            with forward_context(), torch.no_grad():
+                # Compute the pseudo labels.
+                pseudo_labels, label_filter = self.pseudo_labeler(self.model, teacher_input)
+                pseudo_labels_inv = self.augmenter.teacher.reverse_transform(pseudo_labels)
+                label_filter_inv = self.augmenter.teacher.reverse_transform(label_filter)
```
anwai98: Same comment as above (line 484)

torch_em/self_training/mean_teacher.py
```diff
@@ -394,3 +397,321 @@ def _validate_impl(self, forward_context):
...
+            xu = xu.to(self.device, non_blocking=True)
+
+            xu1, xu2 = self.augmenter.teacher.transform(xu), self.augmenter.student.transform(xu)
+            teacher_input, model_input = xu1, xu2
+
+            with forward_context(), torch.no_grad():
+                # Compute the pseudo labels.
+                pseudo_labels, label_filter = self.pseudo_labeler(self.teacher, teacher_input)
+                pseudo_labels_inv = self.augmenter.teacher.reverse_transform(pseudo_labels)
+                label_filter_inv = self.augmenter.teacher.reverse_transform(label_filter)
```
anwai98: Same comment as above in `fix_match.py` (line 484)

And some other similar occurrences below

torch_em/self_training/fix_match.py
```diff
@@ -361,3 +366,318 @@ def _validate_impl(self, forward_context):
...
+            xu1, xu2 = self.augmenter.teacher.transform(xu), self.augmenter.student.transform(xu)
+            teacher_input, model_input = xu1, xu2
+
+            # Perform supervised training.
+            self.optimizer.zero_grad()
+            with forward_context():
+                # We pass the model, the input and the labels to the supervised loss function,
+                # so that how the loss is calculated stays flexible, e.g. to enable ELBO for PUNet.
+                supervised_pred = self.model(xs)
+                supervised_loss = self.supervised_loss(supervised_pred, ys)
```
anwai98, line 552: Just an FYI, it looks like the expected loss input changed here compared to existing self-training trainers. So far, the loss seems to receive something like `loss(model, input_, labels)` or `loss(model, input_, pseudo_labels, label_filter)`, whereas here it is called with the prediction directly, e.g. `loss(prediction, labels)` or `loss(prediction, pseudo_labels, label_filter)`. I just wanted to bring this up that this may affect compatibility with some of the existing losses / metrics, in case we intend to keep two interfaces.

Same comment for line 589 in `mean_teacher.py`

test/transform/test_invertible_augmentations.py
anwai98, line 1: The current tests cover the invertibility of the augmentation wrapper itself quite nicely, but not the new self-training trainer paths that use it. Since the existing FixMatch / MeanTeacher trainers already have dedicated integration tests, it may be helpful to add one small regression test for the new invertible trainer variants as well.

anwai98 (commented):
Thanks a lot for the changes, looks awesome to me @Marei33! I really like the invertible augmentation design and how lightweight the integration into the trainers is. I left a few smaller comments mostly around edge cases and interface consistency, but overall I think this looks great.

test/transform/test_invertible_augmentations.py
constantinpape: I discussed with Marei that we'll not add tests for this right now, as these would increase the test runtimes quite a bit. Maybe we can reconsider this later, but let's keep it out of the PR to keep it limited.

anwai98: Nice, okie!

torch_em/self_training/fix_match.py
Marei33: Yes, you're right. Will check that again

torch_em/self_training/fix_match.py
Marei33: Jaaa, will also change that

torch_em/self_training/fix_match.py
Marei33: Yes, that's true. That was on purpose, because otherwise the augmenter needs to be passed to the loss as well to inverse the model's prediction before computing the loss. So I thought, this way is more flexible...

Marei33:
@anwai98 changed some code lines based on your comments, should be resolved now

anwai98 (approved):
Hi @Marei33,

The changes look great to me. Thanks a ton!

There’s one small robustness improvement we should make later: right now we rely on the augmenter exposing `reset_all`, `teacher`, and `student` implicitly. It would be cleaner to define a common base interface, for example `BaseInvertibleAugmenters`, with these required methods / attributes, let the concrete invertible augmenters inherit from it, and validate against that interface when wiring the trainer. I don’t think this needs to block the PR though, and we can follow up on it later on.

For now, I suggest we go ahead and merge this! @constantinpape

constantinpape:
Thanks for reviewing @anwai98. I went ahead and merged it.

> right now we rely on the augmenter exposing reset_all, teacher, and student implicitly. It would be cleaner to define a common base interface, for example BaseInvertibleAugmenters, with these required methods / attributes, let the concrete invertible augmenters inherit from it, and validate against that interface when wiring the trainer.

This makes sense. But I also think that's better to do in a separate follow-up PR.

anwai98:
> This makes sense. But I also think that's better to do in a separate follow-up PR.

+1 definitely!

PR #667, opened 2026-04-22
Fix multi-channel logging in self-training

### LINKED ISSUE #565: Using images with multiple channels for the MeanTeacherTraine causes an error
When training a model that expects images with multiple (in this case 2) channels, results in an error:

`RuntimeError: stack expects each tensor to be equal size, but got [2, 640, 640] at entry 0 and [1, 640, 640] at entry 2.`

It originates in the logger at line 57, because it tries to stack the images with the predictions and pseudo-labels. 

As a work around, changing lines 49 and 50 to
```
torch_em.transform.raw.normalize(x1[0, 0:1]),
torch_em.transform.raw.normalize(x2[0, 0:1]),
```
just as is done for the predictions and pseudo-labels, makes the code run.

anwai98:
Fixes #565.

Use a single display channel when logging multi-channel self-training inputs so tensorBoard image stacking works with 1-channel predictions and pseudo-labels.

I'm gonna leave this open for now to avoid any other incoming `self_training` merge conflict issues. I'll wait until other PRs are in, and sort this out only then!

### DIFF
--- a/torch_em/self_training/logger.py
+++ b/torch_em/self_training/logger.py
@@ -13,6 +13,10 @@ class SelfTrainingTensorboardLogger(torch_em.trainer.logger_base.TorchEmLogger):
         trainer: The instantiated trainer class.
         save_root: The root directory for saving the checkpoints and logs.
     """
+    @staticmethod
+    def _get_image_channel(x):
+        return x[0, 0:1] if x.shape[1] > 1 else x[0]
+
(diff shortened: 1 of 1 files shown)


anwai98:
This is ready to review now! (took care of the merge conflicts)

Lemme know how it looks @constantinpape

constantinpape (approved):

PR #668, opened 2026-04-22
Handle dataLoaders without explicit shuffle attribute

### LINKED ISSUE #564: MeanTeacherTrainer requires each dataloader to have 'shuffle' attribute, but typical torch Dataloaders do not have this.
When creating MeanTeacherTrainer model trainer, I get the error

`ValueError: <class 'torch_em.self_training.mean_teacher.MeanTeacherTrainer'> requires each dataloader to have 'shuffle' attribute.`

Typically an instance of the torch.utils.data.DataLoader class does not have a "shuffle"-attribute.

The bug is easy to work around, though, by adding the attribute manually, e.g., `train_dataloader.shuffle = True`.

anwai98:
Hi @constantinpape,

This PR aims to fix #564 (but it's a bit tricky since there's a monkey patch which I'm not familiar with, in terms of the design choice for it).

The `DefaultTrainer` checkpoint serialization previously assumed that every `DataLoader` exposed a `.shuffle` attribute. This breaks for external classical PyTorch dataloaders, even though they are valid inputs for training.

This change keeps the existing `.shuffle` convention for loaders created or restored inside `torch-em`, but infers `shuffle` from the loader sampler when saving constructor arguments. This makes trainer checkpoint save / load work for both `torch-em` dataloaders and external PyTorch dataloaders without requiring users to monkey-patch `.shuffle` themselves (as suggested by the user in the issue above)

What do you think about this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ logs/
 *.pt
 *.DS_Store
 .claude/
+.codex
--- a/test/trainer/test_default_trainer.py
+++ b/test/trainer/test_default_trainer.py
@@ -30,14 +30,26 @@ class TestDefaultTrainer(unittest.TestCase):
         rmtree(self.checkpoint_folder, ignore_errors=True)
         rmtree(self.log_folder, ignore_errors=True)
 
-    def _get_kwargs(self, with_roi=False, compile_model=False):
+    def _get_loader(self, with_roi=False, external=False):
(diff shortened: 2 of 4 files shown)


constantinpape:
I am a bit confused about the definition of an external dataset here, I would assume that is not a torch-em dataset, but you do initialize a torch-em dataset in this case. Let's discuss.

constantinpape (approved):

PR #669, opened 2026-04-22
Validate empty ROIs in segmentation dataset

### LINKED ISSUE #274: Error messages for ROIs
## Description
When using ROIs in the torch_em Dataloader (I used: `torch_em.default_segmentation_loader`) there is a generic error message if the ROIs do not fit (e.g. empty or incorrect size).
Error message:
``` 
File "/home/freckmann15/miniforge3/envs/sam/lib/python3.11/site-packages/torch/utils/data/dataloader.py", line 631, in __next__
    data = self._next_data()
           ^^^^^^^^^^^^^^^^^
  File "/home/freckmann15/miniforge3/envs/sam/lib/python3.11/site-packages/torch/utils/data/dataloader.py", line 1319, in _next_data
    raise StopIteration
StopIteration
```
This message is quite difficult to interpret and therefore it is difficult to fix.
The error can be fixed by having a min_shape for the ROIs so ROIs are at least the shape of the patch_shape used.
I would suggest a more specific error message for this case.

anwai98:
This should in principle fix #274.

The issue seemed to be that for invalid empty ROIs in `default_segmentation_loader`, it previously failed with a low-level error from `RoiWrapper`, which made the real problem hard to understand.

This PR aims to add validation for ROIs before wrapping the data and raises a clear `ValueError` when the ROI results in an empty region, i.e. for small but non-empty ROIs should still be allowed and continue to work via padding.

What do we think? @lufre1 @constantinpape

### DIFF
--- a/torch_em/util/util.py
+++ b/torch_em/util/util.py
@@ -102,6 +102,40 @@ def ensure_tensor(tensor: Union[torch.Tensor, ArrayLike], dtype: Optional[str] =
     return tensor
 
 
+def validate_roi(roi, shape, patch_shape=None):
+    """Normalize an ROI to explicit slices and validate that it is non-empty."""
+    if roi is None:
+        return None
+    if isinstance(roi, slice):
+        roi = (roi,)
+    if not isinstance(roi, tuple):
+        raise TypeError(f"Invalid roi type: {type(roi)}")
+    if len(roi) > len(shape):
+        raise ValueError(f"Invalid roi {roi} for data shape {shape}: too many dimensions")
+
+    normalized_roi = []
+    for this_roi, dim in zip(roi, shape):
+        if not isinstance(this_roi, slice):
+            raise TypeError(f"Invalid roi entry: {this_roi}. Only slices are supported")
+        step = 1 if this_roi.step is None else this_roi.step
+        if step != 1:
+            raise ValueError(f"Invalid roi {roi}: slice steps other than 1 are not supported")
+        start, stop, _ = this_roi.indices(dim)
+        normalized_roi.append(slice(start, stop))
+
+    if len(roi) < len(shape):
+        normalized_roi.extend(slice(0, dim) for dim in shape[len(roi):])
+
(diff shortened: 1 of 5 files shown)


torch_em/util/util.py
```diff
@@ -102,6 +102,40 @@ def ensure_tensor(tensor: Union[torch.Tensor, ArrayLike], dtype: Optional[str] =
     return tensor
 
 
+def validate_roi(roi, shape, patch_shape=None):
```
lufre1, line 105: almost waterproof as far as i tested! 
But what about if the ROI is smaller than patch_shape? e.g.
roi = (slice(0, 10), slice(0, 10), slice(0, 10))
patch_shape = (64, 64, 64)

anwai98: > But what about if the ROI is smaller than patch_shape? e.g.
roi = (slice(0, 10), slice(0, 10), slice(0, 10))
patch_shape = (64, 64, 64)

This case is already handled via padding, i.e. the ROI is sampled from the available region, and them padded to `patch_shape` when `with_padding` is set to True (i.e. the default). Just to be sure about it too, I'll add another test for the same scenario explained to be explicit for checking this! ;)

torch_em/util/util.py
```diff
@@ -102,6 +102,40 @@ def ensure_tensor(tensor: Union[torch.Tensor, ArrayLike], dtype: Optional[str] =
...
+        raise ValueError(f"Invalid roi {roi} for data shape {shape}: too many dimensions")
+
+    normalized_roi = []
+    for this_roi, dim in zip(roi, shape):
+        if not isinstance(this_roi, slice):
+            raise TypeError(f"Invalid roi entry: {this_roi}. Only slices are supported")
+        step = 1 if this_roi.step is None else this_roi.step
+        if step != 1:
+            raise ValueError(f"Invalid roi {roi}: slice steps other than 1 are not supported")
+        start, stop, _ = this_roi.indices(dim)
```
constantinpape, line 123: I am a bit confused by this, we can discuss.

PR #670, opened 2026-04-26
Add label path caching for efficient livecell dataloading

anwai98:
Okay this has been annoying me for a while. I decided to take care of it. This makes the label path loading efficient for the livecell data. Otherwise, one has to make the coco-style preparation everytime, and it burns a few minutes before the dataloading works out!

This should work now, and GTM from my side!

### DIFF
--- a/torch_em/data/datasets/light_microscopy/livecell.py
+++ b/torch_em/data/datasets/light_microscopy/livecell.py
@@ -88,10 +88,6 @@ def _annotations_to_instances(coco, image_metadata, category_ids):
     return seg.astype("uint16")
 
 
-def _resolve_cached_paths(cache_dir, paths):
-    return [path if os.path.isabs(path) else os.path.join(cache_dir, path) for path in paths]
-
-
 def _create_segmentations_from_annotations(annotation_file, image_folder, seg_folder, cell_types):
     # Use a per-cell_types cache to avoid reloading the COCO JSON when data is already prepared.
     cache_key = "all" if cell_types is None else "_".join(sorted(cell_types))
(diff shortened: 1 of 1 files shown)


torch_em/data/datasets/light_microscopy/livecell.py
```diff
@@ -87,7 +88,22 @@ def _annotations_to_instances(coco, image_metadata, category_ids):
...
+
+
 def _create_segmentations_from_annotations(annotation_file, image_folder, seg_folder, cell_types):
+    # Use a per-cell_types cache to avoid reloading the COCO JSON when data is already prepared.
+    cache_key = "all" if cell_types is None else "_".join(sorted(cell_types))
+    cache_file = os.path.join(seg_folder, f"seg_paths_{cache_key}.json")
+    if os.path.exists(cache_file):
+        with open(cache_file) as f:
+            cached = json.load(f)
+        cache_dir = os.path.dirname(cache_file)
```
constantinpape: This will always resolve to `seg_folder` and is thus unnecessary. Just leave it at `seg_folder`.

torch_em/data/datasets/light_microscopy/livecell.py
```diff
@@ -87,7 +88,22 @@ def _annotations_to_instances(coco, image_metadata, category_ids):
...
+        image_paths = _resolve_cached_paths(cache_dir, cached["image_paths"])
```
constantinpape: You know that you have relative paths in there, so just write
```
... = [os.path.join(seg_folder, fname) for fname in cached[...]]
```

torch_em/data/datasets/light_microscopy/livecell.py
anwai98: Sorted!

torch_em/data/datasets/light_microscopy/livecell.py
anwai98: Ah indeed true. I think the rationale was just to avoid repeated code, but it isn't relevant since it's always relative paths and we operate using list comprehensions anyways. Thanks for spotting!

constantinpape (approved):

PR #673, opened 2026-04-27
Added storage in .h5 files in a parallelized pipeline

titusgriebel:
Hi @anwai98 
As we discussed, I added a logic that stores images and labels in .h5 files:  
- I also rewrote some of the path logic in order to improve readability. 
- Since the runtime for h5 writing including the gzip compression was quite long, I also added a multiprocessing logic to speed things up. 
- I added a logic that deletes the raw image and label files as well as the .tar.gz file after successful processing
- I tested the code for one cell type and found it to work as intended


anwai98 (approved):
LGTM. Thanks @titusgriebel. I'll go ahead and merge this!

PR #674, opened 2026-04-30
Add BBBC030, Glioma C6, MorphoSeg and IDR0095 datasets

anwai98:
This PR adds 4 (+1) more LM datasets. Most of them are sparsely annotated, but three of them are pretty nice. Also found a DeepBacs data out there from the OG paper, that we were missing.

And made some more linting fixes.

Will go ahead and merge this once the tests pass!


anwai98:
Nice. I'll go ahead and merge this!

PR #675, opened 2026-04-30
Add Orion CRC dataset

anwai98:
There's a bit of a corner case for one of the datasets. I'll debug it later and only merge it then!


anwai98:
Okay this one needed some clean up but it's sorted now. It's a cool paired at for H&E and mIF paired images combined with annotations. GTM from my side!

PR #681, opened 2026-05-08
Fix geopandas df loading.

titusgriebel:
@anwai98 
This PR fixes an issue where JSON-encoded strings were not properly decoded when loading the GeoDataFrame.
Likely due to another change by the data curators, given this script's history?


anwai98:
Hi @titusgriebel,

Thanks for the PR!

> Likely due to another change by the data curators, given this script's history?

This could be the case. Good catch!

Looks good to me on a higher level. I'll go ahead and merge this right away!

PR #685, opened 2026-05-19
Add CEFA Hela dataset

anwai98:
This PR adds a new SBF-SEM dataset for multiple organelles and cells, the only thing is that it is a semantic segmentation dataset, and the raw volumes are massive, so I didn't manage to debug it locally. I'll check it over the cluster once it's back!


anwai98:
Sorted!

PR #687, opened 2026-05-21
Add LICONN dataset

anwai98:
I need to check if all works well with zarrv3 style storage. I checked yesterday, and sharding (with a practical shard-chunk-combination) saves inodes by a lot!

Lemme check how the images and corresponding labels look (I have a vague memory, need to brush it up hehe) and only then I'll consider merging this!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ logs/
 *.DS_Store
 .claude/
 .codex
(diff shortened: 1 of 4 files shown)


anwai98:
Looks pretty neat ;) I'll go ahead and merge this!

PR #689, opened 2026-05-22
Pin python version in docs CI

anwai98:
Just checking something for now! (I think something got sherlocked which is breaking the docs CI)


anwai98:
Yeah I think it happened somehow due to `nifty`'s new build drifting towards weird py314 issues. This should be working now. I'll go ahead and merge this!

PR #690, opened 2026-05-22
Extend cremi data to support padded volumes

anwai98:
This PR adds support for the bigger CREMI volumes, but only allows to switch them in `get_cremi_data` and `get_cremi_paths`. The `_dataset` and `_loader` cannot have access to this in any case (I only want the large volumes to fetch the raw images and run predictions).

@constantinpape what do you think about this?

### DIFF
--- a/torch_em/data/datasets/electron_microscopy/cremi.py
+++ b/torch_em/data/datasets/electron_microscopy/cremi.py
@@ -103,6 +103,10 @@ def get_cremi_paths(
 
     Returns:
         The filepaths to the training data.
+
+    Note:
+        The padded volumes are not available via the dataset and dataloader functions; use this function directly
+        if you need access to the padded data.
     """
     get_cremi_data(path, samples, version, use_realigned, download)
     suffix = "_padded" if version == "padded" else ""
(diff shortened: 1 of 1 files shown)


constantinpape (commented):
I am not sure if this makes so much sense. The raw data and labels for the padded volumes won't be aligned as far as i can tell. So the data loader will not make sense.

anwai98:
> The raw data and labels for the padded volumes won't be aligned as far as i can tell. So the data loader will not make sense.

Users cannot access the padded volumes from `get_cremi_dataset` or `get_cremi_loader`.

The intention was just to have access to the raw data via `get_cremi_paths`. The access to the padded volumes to the loader is kinda blocked (`version` is hard-coded to use the cropped volumes, keeping the state as before)

constantinpape:
I see. That makes sense.

anwai98:
Awesome sauce. Then I'll merge this once the tests pass. Thanks @constantinpape!

constantinpape:
Maybe add a comment on this in the doc-string of `get_cremi_paths`, i.e. explaining that the padded volumes are not available in the functions that yield dataloader / dataset.

anwai98:
> Maybe add a comment on this in the doc-string of get_cremi_paths, i.e. explaining that the padded volumes are not available in the functions that yield dataloader / dataset.

Okie, on it!

PR #691, opened 2026-05-27
Add NISB data

anwai98:
Finally managed to get sometime to implement this hehe.

I'll check if the zarrv3 conversion works out fine first, and only then merge it!


anwai98:
Works like a charm. I'll go ahead and merge this!

PR #696, opened 2026-05-29
Add IGNITE dataset

titusgriebel:
@anwai98 
This PR adds the ignite dataset (ref. [https://github.com/DIAGNijmegen/ignite-data-toolkit](url)) to torch_em.
Although they also offer centroid annotations in IHC images, I decided to only implement H&E with its tissue segmentation for now since only H&E can be used out of the box.


anwai98 (approved):
I pushed a few touch up changes, looks good to me. Thanks @titusgriebel!

I'll go ahead and merge this once the tests pass!

PR #702, opened 2026-05-30
Refactor to use bioimage-cpp instead of vigra, nifty, and affogato

constantinpape:


constantinpape:
@anwai98 the migration to bioimage-cpp is ready now and we are rid of vigra / nifty dependencies.

Once this is good to go from your end I will add CI to publish to PyPI and bump the version.

anwai98 (approved):
I had an overall look at the changes. Looks great to me. Thanks @constantinpape!

Let's go ahead and merge this!

PR #708, opened 2026-06-03
Add Parlakgul and human liver datasets

anwai98:


anwai98:
I'll wait for the CIs. The data looks pretty good, but both are organelles (with semantic annotations). I'll merge this once the tests pass!

PR #718, opened 2026-06-11
Support bfloat16 in mixed precision training

chuyaowang:
The PRs mentioned in the [issue](https://github.com/computational-cell-analytics/micro-sam/issues/1214)

Allows callers to select the autocast dtype for mixed precision training instead of the hardcoded torch.float16. Defaults to torch.float16 to preserve existing behaviour.

- Add `mixed_precision_dtype: torch.dtype = torch.float16` parameter to `DefaultTrainer.__init__` with corresponding docstring entry
- Store `self.mixed_precision_dtype` as an instance attribute so the Serializer can round-trip it through checkpoints
- Create GradScaler only when dtype is float16; bfloat16 has fp32 range and does not require gradient scaling
- Pass `dtype=self.mixed_precision_dtype` to the autocast context in `_train_epoch_mixed` and `_validate_mixed`
- Select `_backprop` (no scaler) vs `_backprop_mixed` (with scaler) based on whether `self.scaler` is None, driven by the chosen dtype

### DIFF
--- a/test/trainer/test_default_trainer.py
+++ b/test/trainer/test_default_trainer.py
@@ -137,6 +137,25 @@ class TestDefaultTrainer(unittest.TestCase):
         self.assertTrue(trainer2.train_loader.shuffle)
         self.assertTrue(trainer2.val_loader.shuffle)
 
+    def test_from_checkpoint_mixed_precision_dtype(self):
+        from torch_em.trainer import DefaultTrainer
+
+        kwargs = self._get_kwargs()
+        kwargs["mixed_precision_dtype"] = "bfloat16"
+        trainer = DefaultTrainer(**kwargs)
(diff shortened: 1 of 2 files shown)


anwai98:
...see response https://github.com/computational-cell-analytics/micro-sam/pull/1234#issuecomment-5250994368

I mention it because this would be the first PR I'd investigate!

anwai98:
Hi @chuyaowang,

I pushed a follow-up to patch `from_checkpoint`, which broke since `torch.dtype` cannot be deserialized from your changes.

All should work now. I'll go ahead and merge this once the tests pass!

anwai98:
And thanks for the PR once again! @chuyaowang 🥳

anwai98 (approved):

PR #719, opened 2026-06-23
Add intensity clipping and custom aug_dict support to invertible augmentations

Marei33:
Extends `torch_em/transform/invertible_augmentations.py` with two features, motivated by self-training experiments where intensity augmentations could push images out of their valid range and where custom augmentation dictionaries were needed.
  
  ## Changes
  
  ### Intensity clipping (`clip_max`)
  - `InvertibleAugmenter` gains a `clip_max` argument. When set, the augmented image is clamped to `[0.0, clip_max]` after the intensity transforms (before the geometrical transforms), keeping it in range.
  - `clip_max` is plumbed through `MeanTeacherAugmenters`, `FixMatchAugmenters`, and `UniMatchv2Augmenters`, which forward it to their `InvertibleAugmenter` instances. Defaults to `None` (no clipping), so existing behaviour is unchanged.
    
  ### Custom augmentation dictionaries (`aug_dict`)
  - Both `get_intensity_augmentations` and `get_geometrical_augmentations` now accept an explicit `aug_dict`, so callers can pass their own augmentation config instead of only the built-in `"weak"` / `"strong"` presets. Either `aug_name` or `aug_dict` must be provided (enforced by an assertion).
    
  ### Cleanups
  - Added `ndim in (2, 3)` assertions to the augmentation factory functions.
  - Fixed copy-pasted/incorrect error messages (these previously reported a dimension error when the real problem was an invalid augmentation name).
  - Renamed the module-private `get_augmentations` helper to `get_default_augmentations` (distinct from the public `torch_em.transform.get_augmentations`, which is unchanged). All internal call sites were updated.
    
  ## Note
  
  `get_intensity_augmentations` / `get_geometrical_augmentations` now take `ndim` as the first positional argument and accept an optional `aug_dict`. All in-tree callers use keyword arguments and are unaffected; external code that passed arguments positionally would need updating.

### DIFF
--- a/torch_em/transform/invertible_augmentations.py
+++ b/torch_em/transform/invertible_augmentations.py
@@ -129,19 +129,21 @@ class InvertibleAugmenter(torch.nn.Module):
         self,
         intensity_transforms: Callable[[torch.Tensor], torch.Tensor],
         geometrical_transforms: Callable[[torch.Tensor], torch.Tensor],
+        clip_max = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
         self.intensity_transforms = intensity_transforms
         self.geometrical_transforms = geometrical_transforms
+        self.clip_max = clip_max
 
     def reset(self):
         self.params = None
 
     def transform(self, x: torch.Tensor) -> torch.Tensor:
-        clip_max = 255.0 if x.max() > 1.0 else 1.0
         x = self.intensity_transforms(x)
-        x = torch.clamp(x, 0.0, clip_max)
+        if self.clip_max is not None:
+            x = torch.clamp(x, 0.0, self.clip_max)
         x = self.geometrical_transforms(x)
 
         self.params = self.geometrical_transforms._params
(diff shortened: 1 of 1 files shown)


Marei33:
@constantinpape ready to review

constantinpape (approved):

constantinpape:
This looks goo from my end. @stmartineau99 could you check if this affects your integration of invertible augmentations in SyanpseNet?

constantinpape:
I went ahead and merged this because Olesia needs it. @stmartineau99 once you have time please check if we need to adjust anything in SynapseNet for it. Thanks!

stmartineau99:
> I went ahead and merged this because Olesia needs it. @stmartineau99 once you have time please check if we need to adjust anything in SynapseNet for it. Thanks!

It seems there is nothing to change, except I will point out that I defined "weak" augmentations dfiferently than Marei. My "weak" augmentations include both intensity and geometric augmentations. 
I edited the DEFAULT_WEAK_AUGMENTATIONS dict so the MeanTeacher will use both, and  I made these augmentations match the behavior of the `weak_augmentations` function previously used by the SynapseNet `get_unsupervised_loader` as closely as possible.

PR #723, opened 2026-07-17
Add RandomPercentileNormalization transform

anwai98:
Here's the augmentation we talked about! What do you think? @constantinpape

### DIFF
--- a/torch_em/transform/raw.py
+++ b/torch_em/transform/raw.py
@@ -140,6 +140,145 @@ def normalize_percentile(
     return normalize(raw, v_lower, v_upper, eps=eps)
 
 
+class RandomPercentileNormalization:
+    """Normalize inputs with randomly sampled percentile bounds.
+
+    By default, the lower percentile is sampled uniformly from ``lower_percentile_bounds``.
+    Normal (Gaussian) sampling can be enabled explicitly with ``distribution="normal"`` and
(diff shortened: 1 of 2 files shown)


torch_em/transform/raw.py
```diff
@@ -140,6 +140,145 @@ def normalize_percentile(
...
+        axis: Axes over which to compute the intensity percentiles.
+        seed: Optional non-negative integer seed for reproducible sampling. NumPy integer types
+            are also supported. Each DataLoader worker derives a distinct stream from this seed.
+            By default, the global NumPy random state is used, which respects DataLoader worker seeding.
+        eps: Epsilon used for numerical stability during normalization.
+    """
+
+    def __init__(
+        self,
+        lower_percentile_bounds: Tuple[float, float] = (0.0, 5.0),
```
constantinpape, line 190: I would give an option to pass `upper_percentile_bounds` too. It can default to `None`, in which case you can use the inverted lower bounds.

constantinpape (commented):
Looks good, only one minor suggestion. Feel free to merge after you implemented it.

torch_em/transform/raw.py
anwai98: Sorted ;) Thanks for the suggestion!

anwai98:
Nice. I'll go ahead and merge this!

PR #725, opened 2026-07-22
Fix puma splits

caroteu:
This makes sure that the data splits for puma all contain each class. Otherwise rare classes tend to disappear in the test or val set. 

cc: @anwai98


anwai98 (approved):
Looks good to me. Thanks @caroteu 

PS. I patched a few other things for consistency.

Will merge this once the tests pass!

PR #726, opened 2026-08-05
Add MyoFuse dataset

anwai98:
This is a very nice data for evaluating instance and semantic segmentation models (because DAPI channel is not available for challenging out-of-myotube nucleus -- there's barely any signal to identify it in myosin channel).

I'll keep this around for future work!


anwai98:
https://www.nature.com/articles/s41598-026-40047-y

PR #727, opened 2026-08-05
Add Pan-Multiplex dataset

anwai98:
It's data from the Nimbus paper. It's quite a lot of rich annotated data. I'll test it out on the cluster and see how much is too much (couldn't download and test it all locally for me)


anwai98:
This looks great. The data is a few hundred gigabytes, took me a couple of hours over the cluster, but the annotations are really mind blowing.

I'll merge this after pushing one final patchfix.

PR #753, opened 2026-08-11
Add NucVerse3d dataset

anwai98:
Pretty rich 3d nucleus segmentation dataset. Will check a few things first (it's a good couple of GBs) and merge it only then!


anwai98:
Sorted!

PR #760, opened 2026-08-17
Update the GitHub Actions to their latest major versions

anwai98:
Moves the first-party actions off the deprecated Node 20 runtime, which GitHub currently force-runs on Node 24. `checkout` v4 -> v7, `setup-python` v5 -> v7, `upload-artifact` and `download-artifact` v4 -> v7.

`download-artifact` stays on the same major as `upload-artifact` instead of going to v8, which is only an ESM migration we do not need.

`setup-micromamba` (v1 in test, v2 in build_docs) and the Pages actions are left for a separate PR, since those can change environment resolution and docs deployment.

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
     - name: Checkout Code
-      uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


anwai98:
Nice, I'll go ahead and merge this!

PR #761, opened 2026-08-21
Make the decoder width of the UNETR models configurable

anwai98:
This helps users gain a bit more control of UNETR (decoder). It's relevant for me because this is a great knob for making the decoder light-weight!

I'll go ahead and merge it (and make a release soon for this too).


anwai98:
PS. The default value is the same, so it should not break for existing methods built on this or `micro-sam`'s offered methods as well!

anwai98:
For future me if I wanna come back to this: there is another scary knob, i.e. the `self.base` engine which plugs the encoder to decoder. We can shrink it too (it gives a nice juicy chunk of parameter gains), but it's very dicy imho since that's the first plug for feature fusion, especially for UNETR3d.

PR #777, opened 2026-09-02
Improve UNETR3D forward with encoder decode split and strip pooling

anwai98:
I found some design and stealth improvements for UNETR3D. This helps mostly to dodge the strip pooling issues on MPS (shipping MPS - CPU device back and forth since it's kernels aren't written in MPS), with the exactly same mathematical outcome. And I went for the encode-decode design so that users in future can extract the UNETR's encoder features directly!

Note: This does not affect any existing users or applications at all. This is just a FYI for everyone pinned! (cc: @caroteu @Marei33 @okorch. And @mugetekin -- this will ease our life for some of your experiments, let's talk details later 😉)

Will go ahead, merge this, and make a new release !

### DIFF
--- a/torch_em/model/unetr.py
+++ b/torch_em/model/unetr.py
@@ -375,6 +375,42 @@ class UNETRBase(nn.Module):
         """
         _check_input_normalization_range(x, expected_range)
 
+    def encode(self, x: torch.Tensor):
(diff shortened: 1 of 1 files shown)


anwai98:
I checked a few more things. All looks safe here. I'll go ahead and merge this!

PR #796, opened 2026-09-04
Map the Platynereis neuropil ids to an ignore label

anwai98:
I think I'd like a review on this, in particular if you could make a fresh download and confirm if the ones I marked as neuropil as actually the case, and if I missed some other neuropil regions too.

cc: @constantinpape


anwai98:
Also gonna merge this once the CIs pass!

